PR #6, opened 2024-01-17
fix: handle request paths starting with //

bajtos:
Calling `new URL('//foo', 'http://127.0.0.1')` produces href `http://foo/`. As a result, we were treating all requests for paths starting with double slash as if no path was specified.

Links:
- https://github.com/filecoin-station/roadmap/issues/57

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -32,7 +32,9 @@ export const createHandler = ({
  * @param {import('pg').Pool} pgPool
  */
 const handler = async (req, res, pgPool) => {
-  const { pathname, searchParams } = new URL(req.url, 'http://127.0.0.1')
+  // Fix the edge case: new URL('//foo', 'http://127.0.0.1') produces href "http://foo/"
+  const reqUrl = req.url.replace(/^\/+/, '/')
+  const { pathname, searchParams } = new URL(reqUrl, 'http://127.0.0.1')
   const segs = pathname.split('/').filter(Boolean)
   if (req.method === 'GET' && segs[0] === 'retrieval-success-rate' && segs.length === 1) {
     await getStatsWithFilterAndCaching(
--- a/test/handler.test.js
+++ b/test/handler.test.js
@@ -51,6 +51,11 @@ describe('HTTP request handler', () => {
     await assertResponseStatus(res, 404)
   })
 
+  it('returns 404 when the path starts with double slash', async () => {
+    const res = await fetch(`${baseUrl}//path-not-found`)
+    await assertResponseStatus(res, 404)
+  })
+
   describe('GET /retrieval-success-rate', () => {
     beforeEach(async () => {
       await pgPool.query('DELETE FROM retrieval_stats')


lib/handler.js
```diff
@@ -32,7 +32,9 @@ export const createHandler = ({
  * @param {import('pg').Pool} pgPool
  */
 const handler = async (req, res, pgPool) => {
-  const { pathname, searchParams } = new URL(req.url, 'http://127.0.0.1')
+  // Fix the edge case: new URL('//foo', 'http://127.0.0.1') produces href "http://foo/"
+  const reqUrl = req.url.replace(/^\/+/, '/')
```
juliangruber, line 36: I disagree with this change. The url `${HOST}//` is invalid and doesn't need to be supported.

bajtos: According to https://stackoverflow.com/a/20524044/69868, which links to several other resources including [RFC 3986 - Uniform Resource Identifier (URI): Generic Syntax](https://datatracker.ietf.org/doc/html/rfc3986), a path starting with a double slash is a valid one.

> Consensus: browsers will do the request as-is, they will not alter the request. The / character is the path separator, but as path segments are defined as:
> ```
> path-abempty  = *( "/" segment )
> segment       = *pchar
> ```
>
> Means the slash after `http://example.com/` can directly be followed by another slash, ad infinitum. Servers might ignore it, but browsers don't, as you have figured out.


TBH, I would rather start using a framework like Fastify that would handle these low-level details for us.

juliangruber: The source says that URLs like this can be valid, but that the server can handle it as it wish. We can be strict and just disallow it. If we want to keep allowing it then I agree that a framework would be great, as what we have now with this PR feels a bit like a hack.

I propose to revert this PR and keep the server simple and strict. Or do you have a use-case for supporting double slashes?

bajtos: Well, I was surprised to learn that a request starting with double slashes was treated the same way as a request starting with `/`. That's a bug I want to fix, to avoid further surprises.

An alternative solution to this PR is to return `404` when `request.url` starts with `//`. Is such a solution in line with your proposal to _be strict and just disallow it_?

```js
if (req.url.startsWith('//')) {
  return notFound(res)
}
```

To be clear, I am happy to change this, but I want to find a solution that behaves as one would expect.

Another option that comes to my mind, I would need to double-check that it works as I think it should work.

```diff
- new URL(req.url, 'http://127.0.0.1')
+ new URL(`http://127.0.0.1${req.url}`)
```

juliangruber: >An alternative solution to this PR is to return 404 when request.url starts with //. Is such a solution in line with your proposal to be strict and just disallow it?

I would like not to mention `//` at all. It's a pity that the URL constructor allows the relative first argument to go "above" the provided base URL. Maybe yeah `new URL(`http://127.0.0.1${req.url}`)` is a simple improvement 👍

PR #8, opened 2024-01-17
refactor: simplify handling of '//' in req.url

bajtos:
Per the discussion in https://github.com/filecoin-station/spark-stats/pull/6#discussion_r1454930123

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -32,9 +32,8 @@ export const createHandler = ({
  * @param {import('pg').Pool} pgPool
  */
 const handler = async (req, res, pgPool) => {
-  // Fix the edge case: new URL('//foo', 'http://127.0.0.1') produces href "http://foo/"
-  const reqUrl = req.url.replace(/^\/+/, '/')
-  const { pathname, searchParams } = new URL(reqUrl, 'http://127.0.0.1')
+  // Caveat! `new URL('//foo', 'http://127.0.0.1')` would produce "http://foo/" - not what we want!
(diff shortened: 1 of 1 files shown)


lib/handler.js
```diff
@@ -32,9 +32,8 @@ export const createHandler = ({
  * @param {import('pg').Pool} pgPool
  */
 const handler = async (req, res, pgPool) => {
-  // Fix the edge case: new URL('//foo', 'http://127.0.0.1') produces href "http://foo/"
-  const reqUrl = req.url.replace(/^\/+/, '/')
-  const { pathname, searchParams } = new URL(reqUrl, 'http://127.0.0.1')
+  // Caveat! `new URL('//foo', 'http://127.0.0.1')` would produce "http://foo/" - not what we want!
```
juliangruber, line 35: What do you think about removing the comment and adding a test for this instead? To me that's the programming native version of this solution

juliangruber (commented):
Otherwise lgtm!

lib/handler.js
bajtos: We already have a test in place, I added it in #6.

https://github.com/filecoin-station/spark-stats/blob/11fc6b1d8f96a9ff855258131620409b3e26ff56/test/handler.test.js#L54-L57

I suppose I can add another test to verify now we handle an existing route with double-slash.

I prefer to keep the comment to remind readers why we are using string concatenation instead of `new URL(path, base)`. In my experience, it's better to be explicit.

juliangruber (approved):

PR #10, opened 2024-01-18
feat: daily & monthly participants

bajtos:
Expose the new public stats introduced by https://github.com/filecoin-station/spark-evaluate/pull/133

Note: the CI will keep failing until we land the spark-evaluate PR and Dependabot upgrades our repo to use the new spark-evaluate main version.

Links:
- https://github.com/filecoin-station/spark-evaluate/issues/79

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -3,7 +3,7 @@ import { json } from 'http-responders'
 import Sentry from '@sentry/node'
 import { URLSearchParams } from 'node:url'
 
-import { fetchRetrievalSuccessRate } from './stats-fetchers.js'
+import { fetchDailyParticipants, fetchMontlyParticipants, fetchRetrievalSuccessRate } from './stats-fetchers.js'
 
 /**
  * @param {object} args
@@ -43,6 +43,20 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchRetrievalSuccessRate)
+  } else if (req.method === 'GET' && segs[0] === 'participants' && segs[1] === 'daily' && segs.length === 2) {
+    await getStatsWithFilterAndCaching(
+      pathname,
+      searchParams,
+      res,
+      pgPool,
+      fetchDailyParticipants)
+  } else if (req.method === 'GET' && segs[0] === 'participants' && segs[1] === 'monthly' && segs.length === 2) {
+    await getStatsWithFilterAndCaching(
+      pathname,
+      searchParams,
+      res,
+      pgPool,
+      fetchMontlyParticipants)
   } else if (req.method === 'GET' && segs.length === 0) {
     // health check - required by Grafana datasources
     res.end('OK')
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -36,9 +36,8 @@ export const createHandler = ({
  * @param {import('pg').Pool} pgPool
  */
 const handler = async (req, res, pgPool) => {
-  // Fix the edge case: new URL('//foo', 'http://127.0.0.1') produces href "http://foo/"
-  const reqUrl = req.url.replace(/^\/+/, '/')
-  const { pathname, searchParams } = new URL(reqUrl, 'http://127.0.0.1')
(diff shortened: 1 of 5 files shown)


lib/handler.js
```diff
@@ -43,6 +47,20 @@ const handler = async (req, res, pgPool) => {
...
       pgPool,
       fetchRetrievalSuccessRate)
+  } else if (req.method === 'GET' && segs[0] === 'participants' && segs[1] === 'daily' && segs.length === 2) {
+    await getStatsWithFilterAndCaching(
+      pathname,
+      searchParams,
+      res,
+      pgPool,
+      fetchDailyParticipants)
+  } else if (req.method === 'GET' && segs[0] === 'participants' && segs[1] === 'monthly' && segs.length === 2) {
```
juliangruber, line 57:
```suggestion
  } else if (req.method === 'GET' && pathname === '/participants/monthly') {
```

no need to use segments if we have static urls

lib/handler.js
```diff
@@ -43,6 +47,20 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchRetrievalSuccessRate)
+  } else if (req.method === 'GET' && segs[0] === 'participants' && segs[1] === 'daily' && segs.length === 2) {
```
juliangruber, line 50:
```suggestion
  } else if (req.method === 'GET' && pathname === '/participants/daily') {
```

lib/handler.js
juliangruber: (also applies above and below)

lib/stats-fetchers.js
```diff
@@ -15,3 +15,45 @@ export const fetchRetrievalSuccessRate = async (pgPool, filter) => {
   }))
   return stats
 }
+
+/**
+ * @param {import('pg').Pool} pgPool
+ * @param {import('./typings').Filter} filter
+ */
+export const fetchDailyParticipants = async (pgPool, filter) => {
+  // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres for converting it into
```
juliangruber, line 24:
```suggestion
  // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres from converting it into
```

lib/stats-fetchers.js
```diff
@@ -15,3 +15,45 @@ export const fetchRetrievalSuccessRate = async (pgPool, filter) => {
...
+  ])
+  return rows
+}
+
+/**
+ * @param {import('pg').Pool} pgPool
+ * @param {import('./typings').Filter} filter
+ */
+export const fetchMonthlyParticipants = async (pgPool, filter) => {
+  // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres for converting it into
```
juliangruber, line 43:
```suggestion
  // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres from converting it into
```

juliangruber (approved):

lib/handler.js
bajtos: Yeah, I was wondering about this too.

But then your suggestion changes how we handle paths like `/participants//monthly` and thus it would introduce inconsistency with the rest of our API 😢 

Maybe I can have a "normalised path string" created via `segments.join('/')` and us that.

WDYT?

juliangruber: If it gets complicated like this, let's just switch to fastify 🤷‍♂️  I don't mind all these edge cases, since I don't yet see any problems arise from them. If you want this to be tighter, I propose let's just make the switch

bajtos: TBH, I don't have an appetite to spend more time on this. I know the current code could be improved; maybe we can also simplify it by not worrying about inconsistencies in edge case handling. Either way, I feel it's a minor detail, and we have bigger fish to fry.

I am going to land what we have now; I added the "Migrate our REST APIs to Fastify" task to our [M4.2 milestone](https://github.com/filecoin-station/roadmap/issues/61).

PR #71, opened 2024-04-12
Feat: Included API handling for fetching from daily_node_metrics

PatrickNercessian:
Links:
filecoin-station/roadmap#96

### DIFF
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -9,7 +9,7 @@ import {
   fetchRetrievalSuccessRate
 } from './stats-fetchers.js'
 
-import { handlePlatformRoute } from './platform-routes.js'
+import { handlePlatformRoutes } from './platform-routes.js'
 
 /**
  * @param {object} args
@@ -69,10 +69,7 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchMinersRSRSummary)
-  } else if (req.method === 'GET' && segs.length === 0) {
-    // health check - required by Grafana datasources
-    res.end('OK')
-  } else if (await handlePlatformRoute(req, res, pgPool)) {
+  } else if (await handlePlatformRoutes(req, res, pgPool)) {
     // no-op, request was handled by handlePlatformRoute
   } else {
     notFound(res)
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -1,7 +1,6 @@
-import assert from 'http-assert'
-import { json } from 'http-responders'
 import Sentry from '@sentry/node'
-import { URLSearchParams } from 'node:url'
+
+import { getStatsWithFilterAndCaching } from './request-helpers.js'
 
 import {
   fetchDailyParticipants,
@@ -10,7 +9,7 @@ import {
   fetchRetrievalSuccessRate
 } from './stats-fetchers.js'
 
-import { fetchDailyNodeMetrics } from './platform-stats-fetchers.js'
+import { handlePlatformRoute } from './platform-routes.js'
 
 /**
  * @param {object} args
@@ -70,16 +69,11 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchMinersRSRSummary)
-  } else if (req.method === 'GET' && segs[0] === 'nodes' && segs[1] === 'daily' && segs.length === 2) {
-    await getStatsWithFilterAndCaching(
-      pathname,
-      searchParams,
-      res,
-      pgPool,
-      fetchDailyNodeMetrics)
   } else if (req.method === 'GET' && segs.length === 0) {
     // health check - required by Grafana datasources
     res.end('OK')
+  } else if (await handlePlatformRoute(req, res, pgPool)) {
+    // no-op, request was handled by handlePlatformRoute
   } else {
     notFound(res)
   }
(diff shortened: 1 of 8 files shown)


juliangruber (approved):

lib/handler.js
```diff
@@ -68,6 +70,13 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchMinersRSRSummary)
+  } else if (req.method === 'GET' && segs[0] === 'nodes' && segs[1] === 'daily' && segs.length === 2) {
+    await getStatsWithFilterAndCaching(
+      pathname,
+      searchParams,
+      res,
+      pgPool,
+      fetchDailyNodeMetrics)
```
bajtos, line 79: We are very aggressively caching the responses from the stats API. Changing the response format requires us to nuke the Cloudflare CDN cache manually. Let's try to get this endpoint right on the first try.

When I look at the SQL query, we return a list of all station IDs seen on a given day.  @PatrickNercessian: How do you envision visualising this data?

I propose to approach the spark-stats API changes from the perspective of the dashboard we are going to build and tailor the data to the needs of those charts.

For example, I can see how it's useful to show how the network size (number of unique Station IDs) evolves over time. I have already implemented such charts for participants (unique wallet addresses). Can we adopt that code for station instances too? (I think it's ok to simply copy&paste. There is also an option to extract shared helpers parameterised by the table name, but that could end up being too complex and not worth the effort right now.) See

- `GET /participants/daily`: 
  https://github.com/filecoin-station/spark-stats/blob/906d6e2788f4ef74389a111c2a6dbdcbbf9106c6/lib/handler.js#L50
- `GET /participants/monthly`: 
  https://github.com/filecoin-station/spark-stats/blob/906d6e2788f4ef74389a111c2a6dbdcbbf9106c6/lib/handler.js#L57
- upcoming `GET /participants/churn-rate`: 
  https://github.com/filecoin-station/spark-stats/pull/74

bajtos (changes requested):
I feel it's important to structure the code implementing these "platform stats" endpoints so that it's easy to mirror future updates from spark-stats to voyager-stats repo.

Let's follow the approach we used in https://github.com/filecoin-station/spark-evaluate/pull/188. 

I.e. there should be two new files - one for the production code and one for the tests - that can be copied as-is from spark-stats to voyager-stats to mirror the changes.

lib/handler.js
PatrickNercessian: > there should be two new files - one for the production code and one for the tests - that can be copied as-is from spark-stats to voyager-stats to mirror the changes.

Makes sense! I did this for platform-stats-fetcher.js already, do you think we should do it for handler code as well? I wasn't sure how we could accomplish that when I was implementing.

> For example, I can see how it's useful to show how the network size (number of unique Station IDs) evolves over time.

I also visualized the data this way, as you are describing it here.

My thought process was to have one endpoint that serves the data needed by Grafana, and Grafana can handle the logic with how it wants to display that data (using Grafana Transformations, which we already plan to use to support multiple Infinity data sources and merge them). This way we don't need to tailor our API to any specific visualizations, and we can just serve the data to Grafana. And we also avoid these cache issues you're describing, since it would instead just be a modification of the Grafana Transformation instead of the underlying API endpoint and its cache. What do you think?

PatrickNercessian:
I need to update this PR to reflect the change from 'metric_date' to 'day' in filecoin-station/spark-evaluate#188

test/handler.test.js
```diff
@@ -244,6 +245,30 @@ describe('HTTP request handler', () => {
       ])
     })
   })
+
+  describe('GET /nodes/daily', () => {
```
bajtos, line 249: Could you please move the new tests into a new test file (e.g. `test/platform-routes.test.js`) so that it's easier to duplicate these tests to Voyager.

lib/handler.js
```diff
@@ -68,6 +70,13 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchMinersRSRSummary)
+  } else if (req.method === 'GET' && segs[0] === 'nodes' && segs[1] === 'daily' && segs.length === 2) {
```
bajtos, line 73: Proposed solution for extracting the platform-specific routes:

```js
  // at the top
  import { handlePlatformRoute } from './platform-routes.js'

  // inside the handler
  if (req.method === 'GET' && segs[0] === 'retrieval-success-rate' && segs.length === 1) {
  // ... (more routes) ...
  } else if (req.method === 'GET' && segs.length === 0) {
    // health check - required by Grafana datasources
    res.end('OK')
  } else if (await handlePlatformRoute(req, res, pgPool) {
    // no-op, request was handled
  } else 
    notFound(res)
  }
```

PatrickNercessian: Sounds good! Would you also prefer that we duplicate the getStatsWithFilterAndCaching function in platform-routes.js?

We could alternatively pass the function in as a parameter to handlePlatformRoute, and when we port to Voyager, we just copy it from handler.js and remove as a handlePlatformRoute parameter.

Up to you!

bajtos: I think `getStatsWithFilterAndCaching` should have the same implementation in Spark and Voyager. In that case, moving the implementation to `platform-routes.js` and letting the project-specific handler file import it from there will require the least amount of maintenance.

Alternatively, and that's an even cleaner option, can we move `getStatsWithFilterAndCaching` and its helpers into its own file?

Let's implement the second option if you agree.

bajtos, line 73: Let's treat the paths `/nodes/daily` and `/nodes/monthly` reserved so that we can implement them to return data in the same format as other `daily` and `monthly` endpoints already do.

We have been already using the term "nodes" to describe unique pairs of `(wallet_address, inet_group)`. Let's use "stations" instead.

Few endpoint names to consider:
- `/stations`
- `/stations/raw`
- `/stations/daily/raw`
- `/stations/daily/full`
- `/stations/daily/details`

PatrickNercessian: Makes sense! Should we also rename the table and references to it in spark-evaluate (it's currently daily_node_metrics)

lib/handler.js
```diff
@@ -10,6 +9,8 @@ import {
   fetchRetrievalSuccessRate
 } from './stats-fetchers.js'
 
+import { handlePlatformRoute } from './platform-routes.js'
```
bajtos, line 12:
```suggestion
import { handlePlatformRoutes } from './platform-routes.js'
```

The number of routes handled by this function should grow over time.

lib/platform-routes.js
```diff
@@ -0,0 +1,24 @@
+import { getStatsWithFilterAndCaching } from './request-helpers.js'
+import { fetchDailyStationMetrics } from './platform-stats-fetchers.js'
+
+export const handlePlatformRoute = async (req, res, pgPool) => {
```
bajtos, line 4:
```suggestion
export const handlePlatformRoutes = async (req, res, pgPool) => {
```

lib/handler.js
```diff
@@ -71,6 +72,8 @@ const handler = async (req, res, pgPool) => {
   } else if (req.method === 'GET' && segs.length === 0) {
     // health check - required by Grafana datasources
     res.end('OK')
```
bajtos, line 74: I like the idea of implementing the health check route as a shared platform route 👍🏻 

Let's remove the duplicate implementation from this file, please.

test/handler.test.js
```diff
@@ -45,6 +47,7 @@ describe('HTTP request handler', () => {
   beforeEach(async () => {
     await pgPool.query('DELETE FROM retrieval_stats')
     await pgPool.query('DELETE FROM daily_participants')
+    await pgPool.query('DELETE FROM daily_node_metrics')
```
bajtos, line 50: I think this cleanup should go to `test/platform-routes.test.js` only, as the tests in this file are not creating any new rows in `daily_node_metrics`. WDYT?

test/platform-routes.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+  beforeEach(async () => {
+    await pgPool.query('DELETE FROM daily_node_metrics')
+  })
+
+  describe('GET /stations/raw', () => {
+    it('returns daily station metrics for the given date range', async () => {
+      await givenDailyStationMetrics(pgPool, '2024-01-10', 'station1')
+      await givenDailyStationMetrics(pgPool, '2024-01-11', 'station2')
+      await givenDailyStationMetrics(pgPool, '2024-01-12', 'station3')
```
bajtos, line 53: Let's enhance the test data to check the following cases:
- Only one station recorded in a day
- More than one station was recorded in a day

```suggestion
      await givenDailyStationMetrics(pgPool, '2024-01-11', 'station2')
      await givenDailyStationMetrics(pgPool, '2024-01-12', 'station2')
      await givenDailyStationMetrics(pgPool, '2024-01-12', 'station3')
```

bajtos (changes requested):
The pull request looks pretty good now! I have a few minor comments to address, PTAL.

test/handler.test.js
PatrickNercessian: Ah yep! I have it in platform-routes.test.js, just forgot to remove it here

lib/handler.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-stats/pull/71/commits/368f30e381bb2434a1a48db942216675e36adaa5 (along with below comments)

bajtos (approved):
LGTM 👍🏻 

I think we need to wait until https://github.com/filecoin-station/spark-evaluate/pull/188 is landed and Dependabot updates the dependency in `package.json`, to get the new DB schema migration scripts that will create the new `daily_stations` table.

https://github.com/filecoin-station/spark-stats/blob/c3ea0b045b94051e77750dd0f201e3c1a75d32f8/package.json#L11

lib/handler.js
bajtos: As we discussed in our call, I am concerned about the egress costs of this solution, but I am happy to try it!

Considering that the new route uses a different path now, allowing us to create `/stations/daily` in the future, if it turns out we cannot use your proposed solution, I am happy to test this in production by landing & deploying the pull request.

PR #74, opened 2024-04-18
feat: participant change rates (monthly)

bajtos:
Implement a new API endpoint `GET /participants/change-rates`

Links:
- https://github.com/filecoin-station/roadmap/issues/91

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@ API exposing public statistics about Spark
 
 Base URL: http://stats.filspark.com/
 
-- `GET /participants/churn-rate?from=<day>&to=<day>`
-  http://stats.filspark.com/participants/churn-rate
+- `GET /participants/change-rate?from=<day>&to=<day>`
+  http://stats.filspark.com/participants/change-rate
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -2,6 +2,13 @@
 
 API exposing public statistics about Spark
 
+## Public API
+
+Base URL: http://stats.filspark.com/
+
+- `GET /participants/churn-rate?from=<day>&to=<day>`
+  http://stats.filspark.com/participants/churn-rate
+
 ## Development
 
 ### Database
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -7,6 +7,7 @@ import {
   fetchDailyParticipants,
   fetchMinersRSRSummary,
   fetchMonthlyParticipants,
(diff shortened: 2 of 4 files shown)


bajtos:
Discussion point:

In the past, I created one endpoint for each chart. I am wondering if I should keep doing the same for churn-rate/growth-rate/retention-rate values because the SQL query and the code to calculate the rates will be very similar for all three rates. Maybe we can have a single endpoint for all three rates?

E.g. `GET /participants/change-rates`:

```json
[
  { "month": "2024-03-01", "churnRate": 0.2, "growthRate": 0.1, "retentionRate": 0.7 }
]
```

@juliangruber WDYT?

juliangruber (approved):
Great work!

juliangruber:
> Discussion point:
> 
> In the past, I created one endpoint for each chart. I am wondering if I should keep doing the same for churn-rate/growth-rate/retention-rate values. The SQL query and the code to calculate the rates will be very similar for all three rates. Maybe we can have a single endpoint for all three rates?
> 
> E.g. `GET /participants/change-rates`:
> 
> ```json
> [
>   { "month": "2024-03-01", "churnRate": 0.2, "growthRate": 0.1, "retentionRate": 0.7 }
> ]
> ```
> 
> @juliangruber WDYT?

I don't have a strong opinion about this. I believe this API is still too young to make a well informed decision. Combining them seems less wasteful, and simpler, so I think it's a good idea for now 👍

bajtos:
> > Discussion point:
> > In the past, I created one endpoint for each chart. I am wondering if I should keep doing the same for churn-rate/growth-rate/retention-rate values. The SQL query and the code to calculate the rates will be very similar for all three rates. Maybe we can have a single endpoint for all three rates?
> > E.g. `GET /participants/change-rates`:
> > ```json
> > [
> >   { "month": "2024-03-01", "churnRate": 0.2, "growthRate": 0.1, "retentionRate": 0.7 }
> > ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @juliangruber WDYT?
> 
> I don't have a strong opinion about this. I believe this API is still too young to make a well informed decision. Combining them seems less wasteful, and simpler, so I think it's a good idea for now 👍

Done 💪🏻 

The pull request is ready for final review and landing.

juliangruber (approved):
great work :)

PR #79, opened 2024-04-22
feat: rename `/miners` to `/storage-providers`

bajtos:
- **feat: rename `/miners` to `/retrieval-providers`**
- **docs: document RSR summary endpoint**

As requested by @PatrickWoodhead in https://space-meridian.slack.com/archives/C06SUTD79NY/p1713361095204799?thread_ts=1713360225.913109&cid=C06SUTD79NY

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,14 @@
 
 API exposing public statistics about Spark
 
+## Public API
+
+Base URL: http://stats.filspark.com/
+
+- `GET /storage-providers/retrieval-success-rate/summary?from=<day>&to=<day>`
+
+  http://stats.filspark.com/storage-providers/retrieval-success-rate/summary
+
 ## Development
 
 ### Database
--- a/lib/handler.js
+++ b/lib/handler.js
@@ -61,7 +61,13 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchMonthlyParticipants)
-  } else if (req.method === 'GET' && segs.join('/') === 'miners/retrieval-success-rate/summary') {
+  } else if (req.method === 'GET' && segs[0] === 'miners') {
+    // Redirect URLs starting with `/miners/` to the new location starting with `/storage-providers/`
+    segs[0] = 'storage-providers'
+    res.setHeader('location', `/${segs.join('/')}?${searchParams}`)
+    res.writeHead(301) // Moved Permanently
+    res.end()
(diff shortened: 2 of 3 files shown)


bajtos:
IMPORTANT:

The response body is still using the `minerId` property. Changing that property name will be a breaking change requiring us to flush all Cloudflare caches.

Alternatively, we can remove the redirect from `/miners` to `/storage-providers` and treat the new path as a new endpoint with a new response body format.

Then, there is a question of when to remove the old `/miners` endpoint. The easiest option is to remove it as part of this pull request. After we land that change, our public dashboard PoC will be broken for a short while until I update the charts to use the new endpoint and the new response body format. IIUC, there is at least one project in the Filecoin ecosystem that is already consuming our Stats API, so they will need to upgrade their code too.

Another aspect to consider: `miner id` is a well-known concept in the Filecoin architecture; the values returned in that property (e.g. `f02099116`) are what people think about as miner IDs. At the same time, the Lotus' `DealProposal` structure stores the miner id in the field called `Provider`, so there is also an argument to support a different field name.

@juliangruber @patrickwoodhead WDYT?

juliangruber:
I'm not familiar with the reasoning for renaming miners to storage providers, especially since it's also common terminology in the Filecoin APIs. Can someone fill me in please?

If we have to rename, I'd do what's simplest without causing service disruption. If we know it's one consumer, let's add `storageProviderId` in addition to `minerId`, and tell them until when `minerId` will still be available. We can also post about this publicly. Not doing the redirect and keeping the old route for backwards compatibility should also work for them, but we need to notify them either way. I think it doesn't make any difference to the consumer which way we choose, and I would go with the least overhead route.

patrickwoodhead:
There was a move a few years ago in the Filecoin ecosystem to use the term `Storage Providers` instead of `Miners` to refer to the Filecoin SPs. I believe this was due to negative connotations around the terms `mining` and `miner`, perhaps due to the wider concerns about how much electricity Bitcoin Proof of Work consumes.

However, given that it is going to be a bit of a hassle to completely remove the term `miner` from the data, and especially given that minerId is a hard coded concept in the Filecoin tech estate, it might actually be more confusing for us to take on both `miner` and `storage provider` in our nomenclature.

Having read the above comments, I propose that we stick with the term `miner` in the API naming and in the code bases, while using `Storage Provider` in blogs, and in the more customer and consumer facing use cases.

bajtos:
Cool, I am closing the PR then.

PR #93, opened 2024-05-06
Fix: Migrate to dedicated station count endpoints

PatrickNercessian:
Roadmap Issue: filecoin-station/roadmap#97

Follow up to https://github.com/filecoin-station/spark-stats/pull/71

### DIFF
--- a/lib/request-helpers.js
+++ b/lib/request-helpers.js
@@ -94,7 +94,7 @@ const setCacheControlForStatsResponse = (res, filter) => {
  * @param {import('./typings').Filter} args.filter
  * @param {string} [args.asColumn]
  */
-export const getDailyDistinctCountQuery = async ({
+export const getDailyDistinctCount = async ({
   pgPool,
   table,
   column,
@@ -126,7 +126,7 @@ export const getDailyDistinctCountQuery = async ({
  * @param {import('./typings').Filter} args.filter
  * @param {string} [args.asColumn]
  */
-export const getMonthlyDistinctCountQuery = async ({
+export const getMonthlyDistinctCount = async ({
   pgPool,
   table,
   column,
--- a/lib/request-helpers.js
+++ b/lib/request-helpers.js
@@ -84,3 +84,30 @@ const setCacheControlForStatsResponse = (res, filter) => {
     res.setHeader('cache-control', `public, max-age=${365 * 24 * 3600}, immutable`)
   }
 }
+
+// Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres from converting it into
+// a JavaScript Date with a timezone, as that could change the date one day forward or back.
+export const getDailyDistinctCountQuery = (table, column, asColumn = null) => {
+  if (!asColumn) asColumn = column + '_count'
+  return `
+    SELECT day::TEXT, COUNT(DISTINCT ${column})::INT as ${asColumn}
+    FROM ${table}
+    WHERE day >= $1 AND day <= $2
+    GROUP BY day
+    ORDER BY day
+  `
+}
+export const getMonthlyDistinctCountQuery = (table, column, asColumn = null) => {
+  if (!asColumn) asColumn = column + '_count'
+  return `
+    SELECT
+      date_trunc('month', day)::DATE::TEXT as month,
+      COUNT(DISTINCT ${column})::INT as ${asColumn}
+    FROM ${table}
+    WHERE
+      day >= date_trunc('month', $1::DATE)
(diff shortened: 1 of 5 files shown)


lib/request-helpers.js
```diff
@@ -84,3 +84,30 @@ const setCacheControlForStatsResponse = (res, filter) => {
...
   }
 }
+
+// Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres from converting it into
+// a JavaScript Date with a timezone, as that could change the date one day forward or back.
+export const getDailyDistinctCountQuery = (table, column, asColumn = null) => {
+  if (!asColumn) asColumn = column + '_count'
+  return `
+    SELECT day::TEXT, COUNT(DISTINCT ${column})::INT as ${asColumn}
+    FROM ${table}
```
juliangruber, line 94: this brings an SQL injection risk, if we're not careful about passing user input into this function. What do you think about instead of return SQL strings here, we also perform the query? The `pg` library will protect us from injections.

```
// platform-stats-fetchers.js
const rows = await getMonthlyDistinctCount({ pgPool, table: 'daily_stations', column: 'statiton_id', from: filter.from, to: filter.to })
```

juliangruber (changes requested):

lib/request-helpers.js
bajtos: I don't expect the table name to be ever provided by the user, but I also agree it's better to be careful.

Let's use [`pg.escapeIdentifier`](https://node-postgres.com/apis/utilities#pgescapeidentifier) to create values that are safe to be included in the SQL query string.

I find it problematic that this function returns the parameterized query but not the parameters. IMO, the function should either return both (query string + parameters array) or execute the query.

PatrickNercessian: Agree with all, addressed in https://github.com/filecoin-station/spark-stats/pull/93/commits/cf9be9cec511e2610598f8a832dde38fbb5f6008

lib/request-helpers.js
```diff
@@ -84,3 +85,72 @@ const setCacheControlForStatsResponse = (res, filter) => {
...
+
+/**
+ * @param {object} args
+ * @param {pg.Pool} args.pgPool
+ * @param {string} args.table
+ * @param {string} args.column
+ * @param {import('./typings').Filter} args.filter
+ * @param {string} [args.asColumn]
+ */
+export const getDailyDistinctCountQuery = async ({
```
bajtos, line 97: Since this function executes the query instead of returning the query, I think it should not have `Query` in the name.

```suggestion
export const getDailyDistinctCount = async ({
```

lib/request-helpers.js
```diff
@@ -84,3 +85,72 @@ const setCacheControlForStatsResponse = (res, filter) => {
...
+
+/**
+ * @param {object} args
+ * @param {pg.Pool} args.pgPool
+ * @param {string} args.table
+ * @param {string} args.column
+ * @param {import('./typings').Filter} args.filter
+ * @param {string} [args.asColumn]
+ */
+export const getMonthlyDistinctCountQuery = async ({
```
bajtos, line 129: Ditto.

```suggestion
export const getMonthlyDistinctCount = async ({
```

bajtos (changes requested):
Almost there!

test/platform-routes.test.js
```diff
@@ -64,9 +64,37 @@ describe('Platform Routes HTTP request handler', () => {
...
+          '/stations/monthly?from=2024-01-11&to=2024-02-11',
+          baseUrl
+        ), {
+          redirect: 'manual'
+        }
+      )
+      await assertResponseStatus(res, 200)
+      const metrics = await res.json()
+      assert.deepStrictEqual(metrics, [
+        { month: '2024-01-01', station_id_count: 3 },
```
bajtos, line 96: I have mixed feelings about the name `station_id_count`. I slightly prefer to use `stations` (we use `participants` in `GET /participants/daily`) or even `count`.

However, I also see how `station_id_count` is a very descriptive & accurate name, so I am fine to keep it.

bajtos (approved):
LGTM 👍🏻 

@juliangruber do you have any more comments?

juliangruber (approved):
Nice! Nothing to add to @bajtos's reviews :)

test/platform-routes.test.js
juliangruber: I like how precise it is, also it is a bit leaky (potentially exposes too many internals). I suggest we iterate on this at a later point, when we understand our system better

PR #102, opened 2024-05-16
Poll FIL events, populate daily_reward_transfers table, reformat endpoint handling

PatrickNercessian:
Related to https://github.com/filecoin-station/spark-evaluate/pull/219

Roadmap links:
* https://github.com/filecoin-station/roadmap/issues/97
* https://github.com/filecoin-station/roadmap/issues/106


bajtos:
@PatrickNercessian I have the new monorepo layout with a new `spark-observer` service almost ready for you, see https://github.com/filecoin-station/spark-stats/pull/120.

There is one more improvement I'd like to implement before I hand this over to you: connect spark-observer to `spark-stats` database and setup a framework for running DB schema migration scripts. I hope to get it done later today.

bin/spark-stats.js
```diff
@@ -50,3 +55,40 @@ console.log('Starting the http server on host %j port %s', HOST, PORT)
...
+  contract: ieContract,
+  provider,
+  rpcUrl: RPC_URL,
+  rpcHeaders
+})
+for await (const event of it) {
+  if (event.name === 'Transfer') {
+    await onTransfer(...event.args)
+  }
+}
```
bajtos, line 94: Please revert changes in this file. The necessary infrastructure should be already prepared for you in https://github.com/filecoin-station/spark-stats/blob/1c069bac5a87de7b2ea58f395bb338f9abe67947/observer/bin/spark-observer.js. Feel free to make any adjustments in that file as needed.

lib/config.js
```diff
@@ -1,3 +1,25 @@
...
+if (RPC_URL.includes('glif')) {
+  rpcHeaders.Authorization = `Bearer ${GLIF_TOKEN}`
+}
+
+export {
+  IE_CONTRACT_ADDRESS,
+  RPC_URL,
+  DATABASE_URL,
+  rpcHeaders
+}
```
bajtos, line 25: It's a good idea to have `lib/config.js`, please move this new code to `observer/lib/config.js` and clean up `observer/bin/spark-observer.js` accordingly.

lib/platform-stats-generator.js
bajtos, line 1: This file should go to `observer/lib/`

test/platform-stats-generator.test.js
bajtos, line 1: This file should go to `observer/test/`

PatrickNercessian:
The functionality of actually checking the blockchain and getting the events is currently untested. I'd like to create a simple dry run script to test it before opening this PR

observer/bin/spark-observer.js
```diff
@@ -62,9 +48,23 @@ await pgPool.query('SELECT 1')
...
+      updateDailyFilStats(
+        pgPool,
+        {
+          to_address: event.args.to,
+          amount: event.args.amount,
+          blockNumber: event.blockNumber
+        }
+      )
+    }
+  })
```
PatrickNercessian, line 70: Also, because I made the switch to use queryFilter instead of "on", this is now a one-time call, instead of a continuously listening action, which is wrong as-is.

We need to either run this script on a schedule (e.g. once daily), or we need to run a schedule within this script's execution)

PatrickNercessian: I think we can do cron jobs using Fly IO config. https://fly.io/docs/machines/flyctl/fly-machine-run/#start-a-machine-on-a-schedule

 What do we think about that?  @juliangruber @bajtos

juliangruber: I'd prefer to run something like

```js
while (true) {
  await run()
  await sleep(ONE_HOUR)
}
```

This way we use less platform features, and own the iteration logic

observer/bin/spark-observer.js
```diff
@@ -1,70 +1,20 @@
...
-  /** @type {number} */
-  const blockNumber = args.pop()
-  console.log('Transfer %s FIL to %s at epoch %s', amount, to, blockNumber)
-  // TODO: update the database
-})
+// Listen for Transfer events from the IE contract
+while (true) {
+  observeTransferEvents(pgPool, ieContract, provider)
+  await new Promise(resolve => setTimeout(resolve, OBSERVATION_INTERVAL_MS))
+}
```
PatrickNercessian, line 20: Should we 'await' observerTransferEvents here?

I think as-is, this means we should run every hour, regardless of how long the observerTransferEvents method takes.

migrations/001.do.sql
```diff
@@ -1 +0,0 @@
-SELECT now();
```
juliangruber, line 1: this migration should be kept, as otherwise the migration library will complain that the hash of already performed migrations doesn't match with what's on disk

migrations/001.do.daily-reward-transfers.sql
```diff
@@ -0,0 +1,11 @@
...
+    day DATE NOT NULL,
+    to_address TEXT NOT NULL,
+    amount NUMERIC NOT NULL,
+    PRIMARY KEY (day, to_address)
+);
+
+CREATE TABLE reward_transfer_last_block (
+    last_block INTEGER NOT NULL
+);
+INSERT INTO reward_transfer_last_block (last_block) VALUES (0);
```
juliangruber, line 11: what about instead we add `block` to `daily_reward_transfers`, and then query that to get the last block seen?

juliangruber: This way we don't run the risk of the two tables getting into an inconsistent state

observer/bin/dry-run.js
```diff
@@ -0,0 +1,24 @@
...
+const fetchRequest = new ethers.FetchRequest(RPC_URL)
+fetchRequest.setHeader('Authorization', rpcHeaders.Authorization || '')
+const provider = new ethers.JsonRpcProvider(fetchRequest, null, { polling: true })
+
+const ieContract = new ethers.Contract(SparkImpactEvaluator.ADDRESS, SparkImpactEvaluator.ABI, provider)
+
+await pgPool.query('DELETE FROM daily_reward_transfers')
+await pgPool.query('DELETE FROM reward_transfer_last_block')
+// Set the last block to -800 to simulate the observer starting from the beginning
+await pgPool.query('INSERT INTO reward_transfer_last_block (last_block) VALUES (-800)')
```
juliangruber, line 20:
```suggestion
```

Can be removed with suggestion above

observer/lib/observer.js
```diff
@@ -0,0 +1,47 @@
...
+ * Observe the transfer events on the Filecoin blockchain
+ * @param {import('pg').Pool} pgPool
+ * @param {import('ethers').Contract} ieContract
+ * @param {import('ethers').Provider} provider
+ */
+export const observeTransferEvents = async (pgPool, ieContract, provider) => {
+  // Get the last checked block. Even though there should be only one row, use MAX just to be safe
+  const lastCheckedBlock = await pgPool.query(
+    'SELECT MAX(last_block) AS last_block FROM reward_transfer_last_block'
+  ).then(res => res.rows[0].last_block)
```
juliangruber, line 13: with suggestion above, needs to be adjusted to select from the transfers table

observer/lib/observer.js
```diff
@@ -0,0 +1,47 @@
...
+      events = await ieContract.queryFilter(ieContract.filters.Transfer(), -1999)
+    } else {
+      throw error
+    }
+  }
+  console.log(`Found ${events.length} Transfer events`)
+  for (const event of events) {
+    const transferEvent = {
+      to_address: event.args.to,
+      amount: event.args.amount
```
juliangruber, line 32: with suggestion above, needs to include `provider.getBlockNumber()`

observer/lib/observer.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Get the current block number and update the last_block in reward_transfer_last_block table
+  // For safety, only update if the new block number is greater than the existing one
+  const blockNumber = await provider.getBlockNumber()
+  console.log('Current block number:', blockNumber)
+  await pgPool.query(`
+    UPDATE reward_transfer_last_block
+    SET last_block = $1
+    WHERE $1 > last_block
+  `, [blockNumber])
```
juliangruber, line 46:
```suggestion
```

observer/bin/spark-observer.js
juliangruber: Awaiting is safer, I think this doesn't need to run exactly every hour. If we want it to be exact and safe, we can await, measure how long it took, and then wait INTERVAL-duration.

juliangruber: We should also handle the case where it errors. I believe a try/catch with console.error() & Sentry.captureException is the least we need to do

migrations/001.do.daily-reward-transfers.sql
PatrickNercessian: We don't actually want the last block with an Event, we want the last "checked" block. So if it's Friday and spark-observer runs, we don't want it to have to check since Sunday (assuming no Transfers since Sunday) for two reasons:

1. We should have already checked all those blocks
2. We will fail because GLIF only allows last 16h40m. Not a big problem bc I have it set to retry, but still doesn't seem like a clean approach.

And I don't see a great way to incorporate what we actually want into daily_reward_transfers. But if you have some specifics in mind for that, I'm open to it!

juliangruber: I'm suggesting to put the last checked block in the above table

> And I don't see a great way to incorporate what we actually want (last "checked" block) into daily_reward_transfers. But if you have some specifics in mind for that, I'm open to it!

Why not put the last checked block in `last_checked_block` in that table?

PatrickNercessian: Oh I think I see, so you mean for every row, we have a new field called `last_checked_block`? Then do we query by MAX to get the highest block seen?

I wonder if it will end up being slow to do that MAX query after we have hundreds of thousands of rows? In the current implementation it's just a single value that is updated over time.

Also, it feels a bit unrelated to each row. Like the `last_checked_block` for any given row doesn't really tell you anything about that row, and we lose some semantic consistency of the table.

That said, I see the benefit of avoiding the `daily_reward_transfers` table not losing consistency with the `last_checked_block` value. Let me know which you prefer

juliangruber: > Oh I think I see, so you mean for every row, we have a new field called `last_checked_block`? Then do we query by MAX to get the highest block seen?

Yeah that's what I'm thinking 👍 

> I wonder if it will end up being slow to do that MAX query after we have hundreds of thousands of rows? In the current implementation it's just a single value that is updated over time.

With an index on the column I don't see performance being an issue. I would rather start correct with this (if it's one table it's consistent by design) and optimize later if necessary.

> Also, it feels a bit unrelated to each row. Like the `last_checked_block` for any given row doesn't really tell you anything about that row, and we lose some semantic consistency of the table.

It tells us at which state of the blockchain this event was received, which is related to the event.

> That said, I see the benefit of avoiding the `daily_reward_transfers` table not losing consistency with the `last_checked_block` value. Let me know which you prefer

I would prefer to fold it into this table. Maybe @bajtos will disagree next week, and if we decide to go that way after all, I think it won't be too bad to follow up again. To me, here, correctness and simplicity outweigh the optimization.

PatrickNercessian: Makes sense! Addressed in https://github.com/filecoin-station/spark-stats/pull/102/commits/3939e8086fa6a0e472515946519f512c76283d07, along with the below suggestions

However, we can't run the dry-run anymore on real events because of the 16h40m max lookback, so we'll have to wait til the next rewards are released (or just merge anyway since it shouldn't break anything else)

juliangruber (changes requested):

juliangruber:
I'm finishing this PR up now, so that I can continue with https://github.com/filecoin-station/spark-stats/pull/131

observer/lib/observer.js
```diff
@@ -0,0 +1,39 @@
...
+  )
+  const lastCheckedBlock = rows[0].last_checked_block
+
+  console.log('Querying impact evaluator Transfer events after block', lastCheckedBlock)
+  let events
+  try {
+    events = await ieContract.queryFilter(ieContract.filters.Transfer(), lastCheckedBlock)
+  } catch (error) {
+    console.error('Error querying impact evaluator Transfer events', error)
+    if (error.message.includes('bad tipset height')) {
```
bajtos, line 21: Relying on specific error message text is fragile. A small change in the error message will break our code.

Is there any machine-readable code we could check to detect this error case?

observer/lib/observer.js
```diff
@@ -0,0 +1,39 @@
...
+      console.log('Block number too old, GLIF only provides last 2000 blocks, querying from -1900')
+      events = await ieContract.queryFilter(ieContract.filters.Transfer(), -1900)
+    } else {
+      throw error
+    }
+  }
+  const currentBlockNumber = await provider.getBlockNumber()
```
bajtos, line 28: Since we are querying the RPC API for the last block number, can we change the queryFilter range logic to be smarter and use the last block number to decide how many blocks we want to query?

Something along the following lines:

```js
const currentBlockNumber = await provider.getBlockNumber()
let queryFrom = lastCheckedBlock
if (queryFrom > currentBlockNumber - 1900) {
  console.log('Block number too old, GLIF only provides last 2000 blocks, querying from -1900')
  queryFrom = currentBlockNumber - 1900
}
```

Then we don't need the fragile check detecting "bad tipset height" error.

Just an idea to consider.

observer/lib/observer.js
```diff
@@ -0,0 +1,39 @@
...
+  console.log('Current block number:', currentBlockNumber)
+  console.log(`Found ${events.length} Transfer events`)
+  for (const event of events) {
+    const transferEvent = {
+      to_address: event.args.to,
```
bajtos, line 33: We use `camelCase` for variables & properties in JavaScript. (I know that's inconsistent with `snake_case` we use for SQL column names.)

Would you mind renaming this property to `toAddress` for consistency with the rest of our (JavaScript) codebase?

observer/test/platform-stats.test.js
```diff
@@ -0,0 +1,71 @@
...
+  })
+
+  let today
+  beforeEach(async () => {
+    await pgPool.query('DELETE FROM daily_reward_transfers')
+
+    // Run all tests inside a transaction to ensure `now()` always returns the same value
+    // See https://dba.stackexchange.com/a/63549/125312
+    // This avoids subtle race conditions when the tests are executed around midnight.
+    await pgPool.query('BEGIN TRANSACTION')
```
bajtos, line 25: There is a subtle bug here: `pgPool.query` grabs a free (or opens a new connection), executes the query, and returns the connection back to the pool.

If we happen to run two queries at the same time, the first will run on the first connection inside this transaction, while the other will run outside of the transaction and therefore it will see different current date.

If you look at the tests in spark-evaluate, they are using a single connection (`pgClient`), and therefore they are not affected by this problem.
https://github.com/filecoin-station/spark-evaluate/blob/main/test/platform-stats.test.js

Here is an example test where we have both `pgPool` and `pgClient`:
https://github.com/filecoin-station/spark-api/blob/a18712008784725f6ebc98eda6a268558ba260b3/api/test/round-tracker.test.js#L24-L33

I propose few alternative ways how to fix this issue:

(1)
Do we really need `pg.Pool` in `updateDailyTransferStats`? Can we rework the code to work accept `pg.Client`? (Note that `pg.Pool` is a superset of `pg.Client`, so you can pass `pg.Pool` to arguments accepting a `pg.Client`).

IIRC, we were using `pg.Pool` in places where we were executing transactions and wanted to avoid a foot gun where a function would assume all queries are executed on the same connection, but the caller would provide a pool, and therefore, some of the queries would be executed on a different connection outside the transaction.

(2)
If we need `pg.Pool,` we need to find a different way to work with the current date. For example, we can call `await getCurrentDate()` in every test. Unfortunately, this is not bullet proof either - if the code under test executes the query just before midnight and the test calls `now()` after midnight, they will see a different day.

(3)
Since this problem manifests only around midnight UTC, I guess we can also ignore it. If we decide so, then let's remove the `BEGIN TRANSACTION` query to avoid a false sense of safety.

@PatrickNercessian @juliangruber WDYT?

bajtos (commented):
Great work! I have a few comments to discuss, see below.

migrations/001.do.daily-reward-transfers.sql
bajtos: While I have concerns about the extra storage the proposed design uses, I also see how it gives us more information when troubleshooting and how it can unlock new possibilities.

I am fine with going down the path you chose.

How can we check this design's impact on DB performance & storage?

observer/lib/observer.js
```diff
@@ -0,0 +1,39 @@
...
+      amount: event.args.amount
+    }
+    console.log('Transfer event:', transferEvent)
+    await updateDailyTransferStats(pgPool, transferEvent, currentBlockNumber)
```
bajtos, line 37: Making one SQL query per participant is expensive. We have ~5k daily participants. Not all of them will receive rewards every week, but still...

We need to check how long does it take to run all these SQL queries.

If it takes too long, then we should rework `observeTransferEvents` to batch SQL queries.

Ideally, we should wait until we collect all events emitted in a single block number, and then run a single SQL query to update all participants that received rewards in that block.

@juliangruber @PatrickNercessian

observer/test/platform-stats.test.js
juliangruber: +1 to (1)

observer/lib/observer.js
juliangruber: +1 to batching, after measuring the impact. Can you run a benchmark with 100k participants and batching vs no batching?

observer/lib/observer.js
PatrickNercessian: Great idea! I also didn't love the error message text check, but the response didn't have any more concrete error code unfortunately.

But I should have thought of this, makes much more sense :)

observer/lib/observer.js
PatrickNercessian: It took ~24.8 seconds to update the table 100,000 times. Tried again with 1,000 updates, and it was 1.4 seconds.

This is of course on my machine, so there may be differences depending on hosting environment. But it seems pretty negligible?

I am open to making the batching change if we prefer it for code quality purposes, even if it doesn't have significant performance gains.

Here was my benchmarking code, if it's of interest:

```
import { updateDailyTransferStats } from '../lib/platform-stats.js'
import { getPgPool } from '../lib/db.js'

const pgPool = await getPgPool()

await pgPool.query('DELETE FROM daily_reward_transfers')

let timeToUpdateTable = 0
for (let index = 0; index < 5_000; index++) {
  const transferEvent = {
    toAddress: "This is fake mock address " + index,
    amount: Math.floor(Math.random() * Number.MAX_SAFE_INTEGER)
  }
  console.log('Transfer event index:', index, 'Transfer event:', transferEvent)
  const beforeTime = Date.now()
  await updateDailyTransferStats(pgPool, transferEvent, 3988923)
  timeToUpdateTable += Date.now() - beforeTime
}

console.log('Time to update table:', timeToUpdateTable)

```
await pgPool.end()

observer/lib/observer.js
PatrickNercessian: Addressed in #134, along with below comments (except batching, awaiting your response about that)

observer/lib/observer.js
juliangruber: Ok thank you for testing this, I believe this is good enough

bajtos: Thank you for measuring the duration, @PatrickNercessian!

I agree with you that 1.4 seconds to update 5000 participants is acceptable.

Could you please ensure we can tell how long this is taking in production? We can keep it super simple - we already log a message before calling the SQL, so let's add another console log after the SQL query is done. Our log service will add timestamps to these messages, and then we can inspect the logs to understand how long it took to observe a weekly payout.

bajtos: We released the rewards earlier today, it seems the entire process of observing the smart contract events finished under one second for ~300 participants. All is good, no code changes are necessary.

PR #131, opened 2024-06-06
Add participant scheduled rewards

juliangruber:
- [x] Expose scheduled rewards
- [ ] Test API
- [ ] Observe scheduled rewards

Blocked by https://github.com/filecoin-station/spark-stats/pull/102

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,10 @@ Base URL: http://stats.filspark.com/
 
   http://stats.filspark.com/participants/change-rates
 
+- `GET /participants/scheduled-rewards/?address=<address>&from=<day>&to=<day>`
+
+  http://stats.filspark.com/participants/scheduled-rewards/
+
 - `GET /miners/retrieval-success-rate/summary?from=<day>&to=<day>`
 
   http://stats.filspark.com/miners/retrieval-success-rate/summary
--- a/stats/lib/typings.d.ts
+++ b/stats/lib/typings.d.ts
@@ -7,4 +7,5 @@ export interface Logger {
 export interface Filter {
   from: string;
   to: string;
+  address?: string;
 }
--- a/migrations/002.do.scheduled-rewards.sql
+++ b/migrations/002.do.scheduled-rewards.sql
@@ -1,0 +1,6 @@
+CREATE TABLE daily_scheduled_rewards (
+  day DATE NOT NULL,
+  address TEXT NOT NULL,
+  scheduled_rewards NUMERIC NOT NULL,
+  PRIMARY KEY (day, address)
+);
--- a/stats/lib/handler.js
+++ b/stats/lib/handler.js
@@ -7,6 +7,7 @@ import {
   fetchMinersRSRSummary,
   fetchMonthlyParticipants,
   fetchParticipantChangeRates,
+  fetchParticipantScheduledRewards,
   fetchRetrievalSuccessRate
 } from './stats-fetchers.js'
 
@@ -70,6 +71,17 @@ const handler = async (req, res, pgPool) => {
       res,
       pgPool,
       fetchParticipantChangeRates)
+  } else if (
+    req.method === 'GET'
+    && segs[0] === 'participants'
+    && segs[1] === 'scheduled-rewards'
+  ) {
+    await getStatsWithFilterAndCaching(
+      pathname,
+      searchParams,
+      res,
+      pgPool,
+      fetchParticipantScheduledRewards)
   } else if (req.method === 'GET' && segs.join('/') === 'miners/retrieval-success-rate/summary') {
     await getStatsWithFilterAndCaching(
       pathname,
--- a/stats/lib/stats-fetchers.js
+++ b/stats/lib/stats-fetchers.js
@@ -113,6 +113,24 @@ export const fetchParticipantChangeRates = async (pgPool, filter) => {
 }
 
 /**
+ * 
+ * @param {import('pg').Pool} pgPool 
+ * @param {import('./typings').Filter} filter 
+ */
+export const fetchParticipantScheduledRewards = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT scheduled_rewards
+    FROM daily_scheduled_rewards
+    WHERE address = $1 AND day >= $2 AND day <= $3
+   `, [
+    filter.address,
+    filter.from,
+    filter.to
+  ])
+  return rows
+}
+
+/**
  * @param {import('pg').Pool} pgPool
  * @param {import('./typings').Filter} filter
  */


juliangruber:
@bajtos do you have an idea how to best test this? I would suggest to pass a mocked IE contract and then query if scheduled rewards have been recorded to the DB.

juliangruber:
Another question is how to get the participant addresses.

When we query the __contract__, we only get those with scheduled rewards over 0.1 FIL. That is insufficient for plotting the road towards the threshold.

Two potential sources for participants are `spark-api`, where we see the addresses of _all_ participants, or `spark-evaluate`, where we see the addresses of all participants _with accepted measurements_. I would like not to add the logic needed for this PR to these repos, because to me, they are not related.  I would like only to add code to `spark-stats`, and _if_ we add code to `spark-api`, it should be generic enough and will let arbitrary services consume this data. No changes to spark-{api,evaluate} should be necessary when in the future we desire to change the semantics of "participants".

Another option is querying the block explorers for `setScores()` calls, which will also include participant addresses _with accepted measurements_. This should work similar to listening for transfer events (https://github.com/filecoin-station/spark-stats/pull/102). However, we are dependent on APIs (like Beryx) to expose this data. Unfortunately, the contract itself can't expose it.

Yet another option is the lazy approach: When a participant's scheduled rewards are requested _for the first time_, we add that participant to the list. This has the upside of least architectural cost, but the downside that you won't see historic rewards from before you first asked for them (read: opened Station Desktop).

Another contract based option is to listen for `MeasurementsAdded` events, then download the raw measurements from w3s, and extract the participant addresses. This should be simple to implement, but is quite clumsy.

Another option is to add an event to the IE smart contract itself, which will post evaluation results to the chain. This comes at the development cost of deploying a new contract.

There are probably other options here too. @bajtos @patrickwoodhead what do you think?

juliangruber:
I think `spark-stats` should consume the smart contract to get this information. That means, either it will...
- listen for `MeasurementsAdded` events, download measurements via IPFS, and extract the participant addresses
- query Beryx for `setScores()` calls, and extract the participant addresses
- listen for a new `ScoresSet` event, and extract the participant addresses, after we deployed a new IE contract with this feature

observer/lib/observer.js
```diff
@@ -37,3 +49,34 @@ export const observeTransferEvents = async (pgPool, ieContract, provider) => {
...
+ * Observe scheduled rewards on the Filecoin blockchain
+ * @param {import('pg').Pool} pgPool
+ * @param {import('ethers').Contract} ieContract
+ */
+const observeScheduledRewards = async (pgPool, ieContract) => {
+  console.log('Querying scheduled rewards from impact evaluator')
+  for (let i = 0; ; i++) {
+    let address
+    try {
+      address = await ieContract.readyForTransfer(i)
```
bajtos, line 63: IIRC, this will not find all participant addresses, only those with more than 0.1 FIL in scheduled rewards.

IMO, the ideal solution is to improve the impact evaluator smart contract to emit a new event whenever the amount of participant's scheduled rewards change. 

Since we don't have that event yet, can we perhaps query the blockchain history to find the `setScores` calls made by spark-evaluate for the last round? The second argument of that function contains a list of participant addresses.

https://github.com/Meridian-IE/impact-evaluator/blob/087c50d401ed7d67de0229bc227f0149764d0671/src/ImpactEvaluator.sol#L118-L122

bajtos: Fortunately, we already have `daily_participants` table that's populated by `spark-evaluate` and that we can access from spark-stats.

I propose to rework the part getting the list of participant addresses to query that table. For example:

```sql
SELECT participant_address
FROM daily_participants
LEFT JOIN participants
ON participant_id = id
WHERE day = now()
```

observer/lib/observer.js
```diff
@@ -37,3 +49,34 @@ export const observeTransferEvents = async (pgPool, ieContract, provider) => {
...
+      continue
+    }
+    console.log('Scheduled rewards for', address, scheduledRewards)
+    await pgPool.query(`
+      INSERT INTO daily_scheduled_rewards (day, address, scheduled_rewards)
+      VALUES (now(), $1, $2)
+      ON CONFLICT (day, address) DO UPDATE SET
+      scheduled_rewards = EXCLUDED.scheduled_rewards
+    `, [address, scheduledRewards])
+  }
```
bajtos, line 81: This is going to be very expensive. With 5k daily active participants (see [Spark Public Dashboard](https://spacemeridian.grafana.net/public-dashboards/32c03ae0d89748e3b08e0f08121caa14?orgId=1)), we are going to make 5k RPC API calls followed by 5k SQL queries.

(1)
Can we group RPC API calls and SQL queries into batches? I believe Ethers v6 supports request batching, and we are configuring the RPC API provider to use batching. Now, we need to find out how to trigger the batching behaviour.

What I would like to see at high level:

```js
const BATCH_SIZE = 10 // for example, we need to tweak this
for (const addrBatch of splitIntoBatches(dailyParticipantAddresses, BATCH_SIZE)) {
  // is this going to trigger Ethers.js batching behaviour?
  const rewards = await Promise.all(
    addrBatch.map(addr => await ieContract.rewardsScheduledFor(addr))
  )

  // run a single SQL query to update multiple rows
  await pgPool.query(`
    INSERT INTO daily_scheduled_rewards (day, address, scheduled_rewards)
    VALUES (now(), unnest($1::text[]), unnest($2::numeric[]))
    ON CONFLICT (day, address) DO UPDATE SET
    scheduled_rewards = EXCLUDED.scheduled_rewards
  `, [
    addrBatch,
    rewards
  ])
}
```

(The query is inspired by an existing query in spark-evaluate [here](https://github.com/filecoin-station/spark-evaluate/blob/bb459b2f2c5b57252a7de659e3827ec5655ec6de/lib/platform-stats.js#L32-L42)).

(2)
We must be mindful of how many requests we send to Glif RPC API. IMO, we shouldn't send all 5k queries as fast as the systems can handle as that would put too much load of the RPC API provider.

I propose to introduce a small delay between iterations of this loop.

If we batch requests for each 10 participants, we need to send ~500 requests. If we use a `1s` delay, then we will finish updating all participants in 500 seconds = 8.3 minutes. I think that's fast enough since the scheduled rewards are updated every 20 minutes.

stats/lib/typings.d.ts
```diff
@@ -7,4 +7,5 @@ export interface Logger {
 export interface Filter {
   from: string;
   to: string;
+  address?: string;
```
bajtos, line 10: This filter is shared by all endpoints. Most endpoints don't support the `address` field.

Can we please create a different filter interface for endpoints that accept an optional `address` field. Also is `address` optional for those endpoints? Judging from your SQL query, this field is actually required.

```ts
export interface DateRangeFilter {
  from: string;
  to: string;
}

export interface AddressFilter {
  address: string;
}
```

Then you can use `DateRangeFilter & AddressFilter` for your new endpoint, and `DateRangeFilter` in all existing endpoints.

bajtos (changes requested):

bajtos:
> @bajtos do you have an idea how to best test this? I would suggest to pass a mocked IE contract and then query if scheduled rewards have been recorded to the DB.

Not really. A mocked IE contract sounds like a good start to me. We already used that approach in voyager-publisher tests IIRC (see https://github.com/filecoin-station/voyager-api/pull/22)

> Another question is how to get the participant addresses.

Great description of different options available 👌🏻 

I posted my thoughts in https://github.com/filecoin-station/spark-stats/pull/131#discussion_r1632224879 before I read your comments.


> I think `spark-stats` should consume the smart contract to get this information. That is the piece that other services should interact with as much as possible, because it lives on the chain and can therefore be trusted.

💯 

Instead of adding `setScores` event, I propose adding an event when participant's scheduled rewards change. There are two cases when that happens - `setScores/ increaseParticipantBalance` increases the balance, `transferScheduled` decreases the balance.

> **Add to spark-api/evaluate**
>
> But how should these services expose the participants, in a way that is not opinionated? (Eg after which period of inactivity should the participant be removed from the list? Should it be queryable by date? Etc).

As I explained in https://github.com/filecoin-station/spark-stats/pull/131#discussion_r1632228589, spark-evaluate maintains the table `daily_participants`. This table is powering our public dashboard, so I think it's safe to assume it will always interpret the term "participant" correctly. 

As for opinions about inactivity & querying by date - I guess the current solution is somewhat opinionated to serve the needs of the dashboard, but I also think it's flexible enough to support additional use cases, e.g. your work in this pull request.

--

Depending on how much time you are willing to spend on this feature, I propose to choose one of the following two ways forward:

- Least effort: Use the existing table `daily_participants`. No changes outside of this PR should be needed.

- Ideal solution: Implement the new smart contract event. It's more work, but the solution may perform better as we can listen to events instead of repeatedly checking the balance of every participant.

observer/lib/observer.js
bajtos: I posted my comment before reading your two comments above.

I replied to your comments here:
https://github.com/filecoin-station/spark-stats/pull/131#issuecomment-2156444081

Sorry for the mess. 🙈

juliangruber:
Awesome, +1 to using `daily_participants` <3

I agree that adding the event is the right solution, but I don't think this is the right time for this

stats/lib/typings.d.ts
bajtos: Started the work here: #137

juliangruber: Will follow up once landed 🙏

db/index.js
```diff
@@ -0,0 +1,75 @@
+import { migrateWithPgClient as migrateEvaluateDB } from 'spark-evaluate/lib/migrate.js'
+import { migrateWithPgClient as migrateStatsDB } from '@filecoin-station/spark-stats-db-migrations'
+import pg from 'pg'
+
+const {
+// DATABASE_URL points to `spark_stats` database managed by this monorepo
+  DATABASE_URL = 'postgres://localhost:5432/spark_stats',
```
bajtos, line 7:
```suggestion
  // DATABASE_URL points to `spark_stats` database managed by this monorepo
  DATABASE_URL = 'postgres://localhost:5432/spark_stats',
```

db/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+const onError = err => {
+  // Prevent crashing the process on idle client errors, the pool will recover
+  // itself. If all connections are lost, the process will still crash.
+  // https://github.com/brianc/node-postgres/issues/1324#issuecomment-308778405
+  console.error('An idle client has experienced an error', err.stack)
+}
+
+export const getStats = async () => {
+  const stats = new pg.Pool({
```
bajtos, line 36: I find it confusing when the name does not mention the database. 

I think it's also important to distinguish between setting up a pool vs. a client.

```suggestion
export const getStatsPool = async () => {
  const statsPool = new pg.Pool({
```

The same comment applies to names related to the "evaluate" DB.

db/index.js
```diff
@@ -0,0 +1,75 @@
...
+  const end = () => Promise.all([stats.end(), evaluate.end()])
+
+  return { stats, evaluate, end }
+}
+
+export const migrate = async () => {
+  const pgPools = await getPgPools()
+
+  console.log('Migrating spark_evaluate database')
+  await migrateEvaluateDB(pgPools.evaluate)
```
bajtos, line 70: Nice, this short name works for me, because it's a property of `pgPools`. 👍🏻

db/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@filecoin-station/spark-stats-db",
```
bajtos, line 2: Hmm, this package does not maintain the database (e.g. the migration scripts), only the client connection factories, right? The name `db` seems a bit misleading then. 

I'd like to propose two options:
- Rename `migrations` package to `db` and merge it with the code introduced in this PR.
- Rename the newly introduced package to something else, e.g. `@filecoin-station/spark-stats-db-clients`

Thoughts?

db/typings.d.ts
```diff
@@ -0,0 +1,6 @@
+import type { Pool } from 'pg'
+
+export interface pgPools {
+  pgPool: Pool;
+  pgPoolEvaluate: Pool;
+}
```
bajtos, line 6: Great to have the types! 💪🏻 

I think you are using different property names in the JS code above.

```suggestion
export interface pgPools {
  stats: Pool;
  evaluate: Pool;
}
```

observer/bin/spark-observer.js
```diff
@@ -3,27 +3,50 @@ import { ethers } from 'ethers'
...
+      } catch (e) {
+        console.error(e)
+        Sentry.captureException(e)
+      }
+      const dt = new Date() - start
+      console.log(`Observing scheduled rewards took ${dt}ms`)
+      await timers.setTimeout((24 * ONE_HOUR) - dt)
+    }
+  })()
+])
```
bajtos, line 52: Would you mind extracting the inner loop functions into standalone functions? I am not a big fan of long nested lambdas.

```js
await Promise.all([
  loopObserveTransferEvents(),
  loopObserveScheduledRewards(),
])
```

observer/lib/observer.js
```diff
@@ -1,25 +1,26 @@
...
  * @param {import('ethers').Provider} provider
  */
-export const observe = async (pgPool, ieContract, provider) => {
+export const observe = async (pgPools, ieContract, provider) => {
   await Promise.all([
-    observeTransferEvents(pgPool, ieContract, provider),
-    observeScheduledRewards(pgPool, ieContract)
+    observeTransferEvents(pgPools.stats, ieContract, provider),
+    observeScheduledRewards(pgPools, ieContract)
   ])
```
bajtos, line 13: What is the purpose of this `observe` method when the file `observer/bin/spark-observer.js` has its own logic to compose these two observers in a different way than presented here? Can we either move that code from `observer/bin/spark-observer.js` to this file, or else remove the `observe()` method entirely?

observer/test/observer.test.js
```diff
@@ -0,0 +1,64 @@
...
+  describe('observeScheduledRewards', () => {
+    let pgPools
+
+    before(async () => {
+      pgPools = await getPgPools()
+    })
+
+    it('observes scheduled rewards', async () => {
+      await pgPools.evaluate.query('DELETE FROM daily_participants')
+      await pgPools.evaluate.query('DELETE FROM participants')
```
bajtos, line 15: Let's move these DB cleanups to a `beforeEach` hook, please.

Tests must not reuse data from other tests, otherwise `it.only` and `mocha -g "<testname>"` won't work.

observer/test/observer.test.js
```diff
@@ -0,0 +1,64 @@
...
+      const { rows: insertRows } = await pgPools.evaluate.query(`
+        INSERT INTO participants
+        (participant_address)
+        VALUES
+          ('0xCURRENT'),
+          ('0xOLD')
+        RETURNING id
+      `)
```
bajtos, line 23: Let's extract this query to a helper function, please. See the existing similar helpers for inspiration:

https://github.com/filecoin-station/spark-stats/blob/831e6694f6515b338afb0ace1c65acdd8cc492c3/stats/test/platform-routes.test.js#L197-L220

https://github.com/filecoin-station/spark-stats/blob/831e6694f6515b338afb0ace1c65acdd8cc492c3/stats/test/handler.test.js#L353-L370

bajtos: In fact, I think you can replace both of these SQL queries with a single call to `givenDailyParticipants()`.

observer/test/observer.test.js
```diff
@@ -0,0 +1,64 @@
...
+        }
+      }
+      await observeScheduledRewards(pgPools, ieContract)
+      const { rows } = await pgPools.stats.query(`
+        SELECT *
+        FROM daily_scheduled_rewards
+      `)
+      assert.strictEqual(rows.length, 1)
+      assert.strictEqual(rows[0].participant_address, '0xCURRENT')
+      assert.strictEqual(rows[0].scheduled_rewards, '100')
```
bajtos, line 48: I think the following assertion will give you more useful details when the assertion fails:

```js
assert.deepStrictEqual([
 { participant_address: '0xCURRENT', scheduled_rewards: '100' }
])
```

If you want to make this test future proof and not fail after we add more fields to the table, then you can limit the columns you are fetching:

```sql
SELECT participant_address, scheduled_rewards
FROM daily_scheduled_rewards
```

observer/test/observer.test.js
```diff
@@ -0,0 +1,64 @@
...
+    })
+    it('updates scheduled rewards', async () => {
+      const ieContract = {
+        rewardsScheduledFor: async () => 200n
+      }
+      await observeScheduledRewards(pgPools, ieContract)
```
bajtos, line 54: IIUC, this test relies on the data created by the previous test. It will fail when executed in isolation.

I propose to initialise the `daily_scheduled_rewards` table via SQL before calling `observeScheduledRewards`.

db/index.js
bajtos: After reading the rest of this PR, I think `getStatsPgPool` would be the best name.

bajtos (changes requested):

observer/test/observer.test.js
juliangruber: sgmt, but how would you propose to share this function between the packages?

observer/lib/observer.js
juliangruber: It's for `dry-run`, so that this file can run all observe functions without needing to know all. Or would you rather inline this logic there?

db/package.json
juliangruber: hm, but what about this?

```
import { migrate } from '@filecoin-station/spark-stats-db-clients'
```

observer/lib/observer.js
juliangruber: move to dry run

observer/test/observer.test.js
juliangruber: move to spark-evaluate

observer/test/observer.test.js
```diff
@@ -10,25 +11,23 @@ describe('observer', () => {
...
-        INSERT INTO daily_participants
-        (participant_id, day)
-        VALUES
-          ($1, now()),
-          ($2, now() - interval '4 days')
-      `, [insertRows[0].id, insertRows[1].id])
+      const today = new Date()
+      await givenDailyParticipants(
+        pgPools.evaluate,
+        `${today.getFullYear()}-${today.getMonth() + 1}-${today.getDate()}`,
```
bajtos, line 20: I believe you can use a Date instance directly; the pg client or the pg server is smart enough to auto-convert the full date-time value to a date.

README.md
```diff
@@ -22,6 +22,10 @@ Base URL: http://stats.filspark.com/
 
   http://stats.filspark.com/participants/change-rates
 
+- `GET /participants/scheduled-rewards/?address=<address>&from=<day>&to=<day>`
+
+  http://stats.filspark.com/participants/scheduled-rewards/
```
bajtos, line 27: The other URLs in this API docs don't use the trailing slash. It would be nice to be consistent, but it's also not a big deal.

observer/bin/migrate.js
```diff
@@ -1,5 +1,3 @@
-import { DATABASE_URL } from '../lib/config.js'
-import { migrateWithPgConfig as migrateStatsDB } from '@filecoin-station/spark-stats-db-migrations'
+import { migrate } from '@filecoin-station/spark-stats-db'
 
-console.log('Migrating spark_stats database')
-await migrateStatsDB({ connectionString: DATABASE_URL })
+await migrate()
```
bajtos: I think now that the observer needs to access the spark-evaluate database, the migration CLI tool should migrate the evaluate database too. Otherwise the following workflow no longer works: 

```
npm run migrate -w observer && npm start -w observer
```

observer/lib/observer.js
```diff
@@ -1,13 +1,14 @@
 import { updateDailyTransferStats } from './platform-stats.js'
+import * as Sentry from '@sentry/node'
```
bajtos, line 2: FWIW, AFAICT, the observer bin script is not initalising Sentry - see `observer/bin/spark-observer.js`.

Fixing that is out of the scope of this pull request, though.

observer/package.json
```diff
@@ -3,18 +3,20 @@
...
   "scripts": {
+    "dry-run": "node bin/dry-run.js",
     "migrate": "node bin/migrate.js",
     "start": "node bin/spark-observer.js",
     "lint": "standard",
     "test": "mocha"
   },
   "devDependencies": {
     "mocha": "^10.4.0",
+    "spark-evaluate": "github:filecoin-station/spark-evaluate#534454c9b2323b67d50b159a3504de8a06eedcb2",
```
bajtos, line 14: I think we should use `#main` in the package json file. Unless you are certain that Dependabot will bump up the commit sha for us.

stats/lib/typings.d.ts
bajtos: @juliangruber This is not resolved yet. Please revert the change in this interface.

In my PR, I decided to describe the endpoint-specific filter properties inline in the "fetcher" definition.

```js
/**
 * @param {import('./typings').DateRangeFilter & {nonZero?: 'true'}} filter
 */
export const fetchRetrievalSuccessRate = async (pgPool, filter) => {
 // ...
}
```

observer/bin/dry-run.js
```diff
@@ -2,8 +2,8 @@ import * as SparkImpactEvaluator from '@filecoin-station/spark-impact-evaluator'
 import { ethers } from 'ethers'
 
 import { RPC_URL, rpcHeaders } from '../lib/config.js'
-import { observe } from '../lib/observer.js'
-import { getPgPools } from '../../db/index.js'
+import { observeTransferEvents, observeScheduledRewards } from '../lib/observer.js'
+import { getPgPools } from '../../db-clients/index.js'
```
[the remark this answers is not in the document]
juliangruber: see follow up commit

observer/test/observer.test.js
juliangruber: should we adjust the type then, to accept string|Date?

README.md
```diff
@@ -24,7 +24,7 @@ Base URL: http://stats.filspark.com/
 
 - `GET /participants/scheduled-rewards/?address=<address>&from=<day>&to=<day>`
```
bajtos, line 25:
```suggestion
- `GET /participants/scheduled-rewards?address=<address>&from=<day>&to=<day>`
```

observer/lib/observer.js
juliangruber: Added it anyway 👼

observer/bin/migrate.js
juliangruber: it's migrating both 👍

PR #137, opened 2024-06-11
refactor: rename Filter to DateRangeFilter

bajtos:
See https://github.com/filecoin-station/spark-stats/pull/131#discussion_r1632233070

### DIFF
--- a/stats/lib/request-helpers.js
+++ b/stats/lib/request-helpers.js
@@ -12,7 +12,7 @@ export const today = () => getDayAsISOString(new Date())
  * @param {URLSearchParams} searchParams
  * @param {import('node:http').ServerResponse} res
  * @param {import('pg').Pool} pgPool
- * @param {(import('pg').Pool, import('./typings').Filter) => Promise<object[]>} fetchStatsFn
+ * @param {(import('pg').Pool, import('./typings.d.ts').DateRangeFilter) => Promise<object[]>} fetchStatsFn
  */
 export const getStatsWithFilterAndCaching = async (pathname, searchParams, res, pgPool, fetchStatsFn) => {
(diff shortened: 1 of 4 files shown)


stats/lib/request-helpers.js
```diff
@@ -12,7 +12,7 @@ export const today = () => getDayAsISOString(new Date())
  * @param {URLSearchParams} searchParams
  * @param {import('node:http').ServerResponse} res
  * @param {import('pg').Pool} pgPool
- * @param {(import('pg').Pool, import('./typings').Filter) => Promise<object[]>} fetchStatsFn
+ * @param {(import('pg').Pool, import('./typings.d.ts').DateRangeFilter) => Promise<object[]>} fetchStatsFn
```
juliangruber, line 15:
```suggestion
 * @param {(import('pg').Pool, import('./typings').DateRangeFilter) => Promise<object[]>} fetchStatsFn
```

juliangruber (approved):

stats/lib/request-helpers.js
bajtos: We would have to revert your change when we add proper `tsc` type checking. Because we use ESM imports in our JS code, TypeScript will require us to use ESM imports in typings too. ESM imports require full file names including the extensions, and TypeScript does the same.

See e.g. https://github.com/filecoin-station/spark-evaluate/pull/252

juliangruber: Should we update the other type imports to include the extension then?

bajtos: > Should we update the other type imports to include the extension then?

Yes, eventually. So far, I have done it as part of enabling type-checking.

PR #138, opened 2024-06-11
fix: last_checked_block retrieval

PatrickNercessian:


observer/lib/observer.js
juliangruber, line 1: fix lgtm, could you please also add a test for this? You can mock `ieContract` so we only test the sql logic

observer/bin/dry-run.js
```diff
@@ -18,4 +18,8 @@ await pgPool.query('DELETE FROM daily_reward_transfers')
 
 await observeTransferEvents(pgPool, ieContract, provider)
 
+// Do it a second time, without clearing the table.
+// This should find 0 events, unless rewards are currently being released.
+await observeTransferEvents(pgPool, ieContract, provider)
```
juliangruber, line 23: I think it's safer to add an automated test instead

juliangruber (changes requested):

observer/bin/dry-run.js
```diff
@@ -20,4 +20,8 @@ await Promise.all([
   observeScheduledRewards(pgPools, ieContract)
 ])
 
-await pgPools.end()
+// Do it a second time, without clearing the table.
+// This should find 0 events, unless rewards are currently being released.
+await observeTransferEvents(pgPools.stats, ieContract, provider)
```
juliangruber, line 25:
```suggestion
```

already covered by tests

juliangruber (approved):

observer/bin/dry-run.js
```diff
@@ -20,4 +20,5 @@ await Promise.all([
   observeScheduledRewards(pgPools, ieContract)
 ])
 
-await pgPools.end()
+
```
juliangruber, line 23:
```suggestion
```

sorry

juliangruber (approved):

observer/lib/observer.js
```diff
@@ -9,9 +9,9 @@ import * as Sentry from '@sentry/node'
  */
 export const observeTransferEvents = async (pgPoolStats, ieContract, provider) => {
   const { rows } = await pgPoolStats.query(
-    'SELECT MAX(last_checked_block) FROM daily_reward_transfers'
+    'SELECT MAX(last_checked_block) AS last_checked_block FROM daily_reward_transfers'
   )
-  let queryFromBlock = rows[0].last_checked_block
+  let queryFromBlock = rows[0].last_checked_block + 1
```
bajtos, line 14: I think there is bug lurking here. 

In the initial run against an empty database, `last_checked_block` will be `undefined` and `queryFromBlock` will be set to `NaN`.

```
❯ node          
Welcome to Node.js v20.14.0.
Type ".help" for more information.
> undefined + 1
NaN
```

juliangruber: You're right!

observer/test/observer.test.js
```diff
@@ -1,19 +1,132 @@
...
-const getDayAsISOString = d => d.toISOString().split('T')[0]
-const today = () => getDayAsISOString(new Date())
+import { observeTransferEvents, observeScheduledRewards } from '../lib/observer.js'
 
 describe('observer', () => {
-  describe('observeScheduledRewards', () => {
-    let pgPools
+  let pgPools
+  const getDayAsISOString = d => d.toISOString().split('T')[0]
+  const today = () => getDayAsISOString(new Date())
```
bajtos, line 11: Was there any particular reason for moving these two helper functions into a `describe` block? Can we move them back to the top where they were before?

juliangruber: You're right here as well

observer/lib/observer.js
PatrickNercessian: We check the truthy value of the queryFromBlock directly after this, so it ends up being reset to the -1900 block

observer/test/observer.test.js
juliangruber: yes please

PR #140, opened 2024-06-12
Add migrate release command

juliangruber:
https://fly.io/docs/reference/configuration/#run-one-off-commands-before-releasing-a-deployment


stats/bin/migrate.js
```diff
@@ -1,12 +1,8 @@
 import {
-  DATABASE_URL,
-  EVALUATE_DB_URL
+  DATABASE_URL
 } from '../lib/config.js'
-import { migrateWithPgConfig as migrateEvaluateDB } from 'spark-evaluate/lib/migrate.js'
 import { migrateWithPgConfig as migrateStatsDB } from '@filecoin-station/spark-stats-db-migrations'
 
-console.log('Migrating spark_evaluate database', EVALUATE_DB_URL)
-await migrateEvaluateDB({ connectionString: EVALUATE_DB_URL })
```
juliangruber, line 9: In prod, this shouldn't try to migrate the evaluation db, which is connected to read only

stats/fly.toml
```diff
@@ -36,3 +36,6 @@ processes = []
   [[services.ports]]
     handlers = ["tls", "http"]
     port = 443
+
+[deploy]
+  release_command = "npm run migrate"
```
bajtos, line 41: IMO, the stats REST API server should not run the migration in production, because it does not own the database schema (does not write new data). The schema is owned by spark-observer.

bajtos (changes requested):
This is a great start!

We need to think a bit more about the right think to do in these three cases:
1. deploying to prod
2. running tests
3. starting the service locally

stats/bin/migrate.js
juliangruber: since we removed `npm run migrate` from the stats production setup, I added evaluation db migration back here

observer/fly.toml
```diff
@@ -7,3 +7,4 @@ primary_region = "cdg"
 
 [deploy]
   strategy = "rolling"
+  release_command = "npm run migrate"
```
bajtos, line 10: I think this will run the root-level migrate script, which in turn runs migrate scripts in all sub-packages (stats, observer). If I am correct, then we can fix that as follows:

```suggestion
  release_command = "npm run migrate --workspace observer"
```

bajtos (changes requested):
IIUC, when spark-observer starts, it calls database migration scripts. So now we run the migration twice - once by Fly.io, second time from our Node.js bin script. Are we okay with that?

https://github.com/filecoin-station/spark-stats/blob/60666900b3d93a8b0d890ad880ac37d2e6591081/observer/bin/spark-observer.js#L14

https://github.com/filecoin-station/spark-stats/blob/60666900b3d93a8b0d890ad880ac37d2e6591081/db/index.js#L71-L73

https://github.com/filecoin-station/spark-stats/blob/60666900b3d93a8b0d890ad880ac37d2e6591081/db/index.js#L48-L54

I propose to modify `getStatsPgPool()` to NOT run the migrations and let the callers handle that aspect.

WDYT?

observer/fly.toml
```diff
@@ -7,3 +7,4 @@ primary_region = "cdg"
 
 [deploy]
   strategy = "rolling"
+  release_command = "npm run migrate --workspace observer"
```
bajtos, line 10: This will migrate both our DB (spark_stats) and the spark-evaluate DB. That does not seem right to me. WDYT?

https://github.com/filecoin-station/spark-stats/blob/e056ce21f6f355555806d31cd0290a4e5bd45f07/observer/bin/migrate.js#L3-L5

juliangruber: You're right! What about we remove `migrateEvaluateDB()` from this bin file?

bajtos: I think that's going to break the second & third use case, won't it?

https://github.com/filecoin-station/spark-stats/pull/140#pullrequestreview-2112812928
> 1. deploying to prod
> 2. running tests
> 3. starting the service locally

How about adding a new `bin/migrate.js` script to the `db` sub-package?

juliangruber: good idea!

bajtos (approved):
:shipit:

PR #144, opened 2024-06-13
chore: setup TypeScript and fix type errors

bajtos:
See https://github.com/space-meridian/roadmap/issues/80

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -72,7 +72,7 @@ export const getEvaluatePgPool = async () => {
 export const getPgPools = async () => {
   const stats = await getStatsPgPool()
   const evaluate = await getEvaluatePgPool()
-  const end = () => Promise.all([stats.end(), evaluate.end()])
+  const end = async () => { await Promise.all([stats.end(), evaluate.end()]) }
 
   return { stats, evaluate, end }
 }
--- a/db/index.js
+++ b/db/index.js
@@ -4,6 +4,9 @@ import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import Postgrator from 'postgrator'
 
+/** @typedef {import('./typings.js').PgPools} PgPools */
+/** @typedef {import('./typings.js').EndablePgPools} EndablePgPools */
+
 export { migrateEvaluateDB }
 
 const {
@@ -62,18 +65,18 @@ export const getEvaluatePgPool = async () => {
 }
 
 /**
- * @returns {Promise<import('./typings').pgPools>}
+ * @returns {Promise<EndablePgPools>}
  */
 export const getPgPools = async () => {
   const stats = await getStatsPgPool()
   const evaluate = await getEvaluatePgPool()
-  const end = () => Promise.all([stats.end(), evaluate.end()])
+  const end = async () => Promise.all([stats.end(), evaluate.end()])
 
   return { stats, evaluate, end }
 }
 
 /**
- * @param {pg.Client} client
+ * @param {pg.Client | pg.Pool} client
  */
 export const migrateStatsDB = async (client) => {
   const postgrator = new Postgrator({
--- a/db/index.js
+++ b/db/index.js
@@ -70,7 +70,7 @@ export const getEvaluatePgPool = async () => {
 export const getPgPools = async () => {
   const stats = await getStatsPgPool()
   const evaluate = await getEvaluatePgPool()
-  const end = async () => Promise.all([stats.end(), evaluate.end()])
+  const end = () => Promise.all([stats.end(), evaluate.end()])
 
   return { stats, evaluate, end }
 }
--- a/db/index.js
+++ b/db/index.js
@@ -4,8 +4,10 @@ import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import Postgrator from 'postgrator'
 
+// re-export types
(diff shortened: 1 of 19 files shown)


db/index.js
```diff
@@ -73,7 +76,7 @@ export const getPgPools = async () => {
 }
 
 /**
- * @param {pg.Client} client
+ * @param {pg.Client | pg.Pool} client
```
bajtos, line 79: Unfortunately, in node-pg typings describe Client and Pool in such way that one cannot be assigned to other 😢

db/typings.d.ts
```diff
@@ -1,6 +1,11 @@
 import type { Pool } from 'pg'
 
-export interface pgPools {
+export interface PgPools {
   stats: Pool;
   evaluate: Pool;
 }
+
+export interface EndablePgPools extends PgPools {
+  end(): Promise<unknown>
```
bajtos, line 9: Our implementation returns `Promise<[void, void]>`. Is it worth changing the implementation so that this line can describe the return value as `void`?

```suggestion
  end(): Promise<void>
```

db/index.js
```diff
@@ -4,6 +4,9 @@ import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import Postgrator from 'postgrator'
 
+/** @typedef {import('./typings.js').PgPools} PgPools */
```
bajtos, line 7: The convention is to use PascalCase for type names, i.e. start with a capital letter. I renamed `pgPools` type to `PgPools`.

db/index.js
juliangruber: What about we add this type to this repo instead?

db/typings.d.ts
juliangruber: sgtm! should be a simple change

observer/lib/platform-stats.js
```diff
@@ -1,5 +1,5 @@
 /**
- * @param {import('pg').Client} pgClient
+ * @param {import('pg').ClientBase} pgClient
```
juliangruber, line 2: for my understanding, why was this change necessary?

stats/bin/spark-stats.js
```diff
@@ -20,6 +20,6 @@ const logger = {
 const handler = createHandler({ pgPools, logger })
 const server = http.createServer(handler)
 console.log('Starting the http server on host %j port %s', HOST, PORT)
-server.listen(PORT, HOST)
+server.listen(Number(PORT), HOST)
```
juliangruber, line 23: what is this cast to number for?

stats/test/handler.test.js
```diff
@@ -38,7 +38,7 @@ describe('HTTP request handler', () => {
     server = http.createServer(handler)
     server.listen()
     await once(server, 'listening')
-    baseUrl = `http://127.0.0.1:${server.address().port}`
+    baseUrl = `http://127.0.0.1:${getPort(server)}`
```
juliangruber, line 41: could we instead do this?
```suggestion
    baseUrl = `http://127.0.0.1:${server.address()!.port}`
```

stats/test/platform-routes.test.js
```diff
@@ -31,7 +31,7 @@ describe('Platform Routes HTTP request handler', () => {
     server = http.createServer(handler)
     server.listen()
     await once(server, 'listening')
-    baseUrl = `http://127.0.0.1:${server.address().port}`
+    baseUrl = `http://127.0.0.1:${getPort(server)}`
```
juliangruber, line 34: as above

juliangruber (changes requested):
massive work 👏

observer/lib/platform-stats.js
bajtos: I'll revert this change. Good catch!

I was experimenting with alternatives to the missing `Queryable` interface (see https://github.com/filecoin-station/spark-stats/pull/144#discussion_r1637976607) and forgot to revert this place.

stats/bin/spark-stats.js
bajtos: The way how TypeScript describes the `listen` method, it accepts only numbers as the port. When we use `PORT`, which can be a string, TypeScript complains.

```js
server.listen('8080', 'localhost')
```

stats/test/handler.test.js
bajtos: Unfortunately not. 

`server.address()` could be either a string (when listening on a named pipe(?)) or an object with a `port` property (when listening over TCP/IP).

We know that `address` will always be an object, but TypeScript cannot infer that.

It's annoying.

db/index.js
bajtos: Hang on! Maybe this will work - let me test it:
```ts
export type Queryable = Pick<Pool, 'query'>
```

bajtos: Works like a charm! See [0324c01](https://github.com/filecoin-station/spark-stats/pull/144/commits/0324c010c4f6817313c12f42b56a3b5e44eda1e9)

db/typings.d.ts
bajtos: Done in [4741c76](https://github.com/filecoin-station/spark-stats/pull/144/commits/4741c7660a1c230d45d1e5602e862c141511d427)

juliangruber (approved):
😍

PR #146, opened 2024-06-13
feat: `GET /deals/daily`

bajtos:
Expose the stats collected as of https://github.com/filecoin-station/spark-evaluate/pull/256

Links:
- https://github.com/space-meridian/roadmap/issues/57

### DIFF
--- a/stats/lib/stats-fetchers.js
+++ b/stats/lib/stats-fetchers.js
@@ -26,6 +26,31 @@ export const fetchRetrievalSuccessRate = async (pgPools, filter) => {
   return stats
 }
 
+/**
+ * @param {import('@filecoin-station/spark-stats-db').pgPools} pgPools
+ * @param {import('./typings').DateRangeFilter} filter
+ */
+export const fetchDailyDealStats = async (pgPools, filter) => {
+  // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres for converting it into
+  // a JavaScript Date with a timezone, as that could change the date one day forward or back.
+  const { rows } = await pgPools.evaluate.query(`
+    SELECT day::text, total, indexed, retrievable
+    FROM daily_deals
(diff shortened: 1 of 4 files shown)


stats/lib/stats-fetchers.js
```diff
@@ -26,6 +26,31 @@ export const fetchRetrievalSuccessRate = async (pgPools, filter) => {
   return stats
 }
 
+/**
+ * @param {import('@filecoin-station/spark-stats-db').pgPools} pgPools
+ * @param {import('./typings').DateRangeFilter} filter
+ */
+export const fetchDailyDealStats = async (pgPools, filter) => {
+  // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres for converting it into
```
juliangruber, line 34:
```suggestion
  // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres from converting it into
```

stats/lib/stats-fetchers.js
```diff
@@ -26,6 +26,31 @@ export const fetchRetrievalSuccessRate = async (pgPools, filter) => {
...
+    `, [
+    filter.from,
+    filter.to
+  ])
+  const stats = rows.map(r => ({
+    day: r.day,
+    total: r.total,
+    indexed: r.indexed,
+    retrievable: r.retrievable
+  }))
```
juliangruber, line 50: I assume you added this out of convention, because it looks like a noop?

juliangruber (approved):

stats/lib/stats-fetchers.js
bajtos: Oh, good point. I was copy-pasting the implementation and did not realise that I don't need to map the rows at all.

I am adding a commit to remove this mapping and directly `return rows`.

PR #142, opened 2024-06-12
Fix daily scheduled rewards api

juliangruber:

PR #147, opened 2024-06-13
Add scheduled rewards handler test

juliangruber:
For https://github.com/filecoin-station/spark-stats/pull/142

### DIFF
--- a/stats/test/handler.test.js
+++ b/stats/test/handler.test.js
@@ -409,10 +409,10 @@ describe('HTTP request handler', () => {
 
   describe('GET /deals/daily', () => {
     it('returns daily deal stats for the given date range', async () => {
-      await givenDailyDealStats(pgPool, { day: '2024-01-10', total: 10, indexed: 5, retrievable: 1 })
-      await givenDailyDealStats(pgPool, { day: '2024-01-11', total: 20, indexed: 6, retrievable: 2 })
-      await givenDailyDealStats(pgPool, { day: '2024-01-12', total: 30, indexed: 7, retrievable: 3 })
-      await givenDailyDealStats(pgPool, { day: '2024-01-13', total: 40, indexed: 8, retrievable: 4 })
+      await givenDailyDealStats(pgPools.evaluate, { day: '2024-01-10', total: 10, indexed: 5, retrievable: 1 })
+      await givenDailyDealStats(pgPools.evaluate, { day: '2024-01-11', total: 20, indexed: 6, retrievable: 2 })
+      await givenDailyDealStats(pgPools.evaluate, { day: '2024-01-12', total: 30, indexed: 7, retrievable: 3 })
+      await givenDailyDealStats(pgPools.evaluate, { day: '2024-01-13', total: 40, indexed: 8, retrievable: 4 })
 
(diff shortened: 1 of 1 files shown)


stats/test/handler.test.js
```diff
@@ -353,17 +350,40 @@ describe('HTTP request handler', () => {
...
+  describe('GET /participants/scheduled-rewards/daily', () => {
+    it('returns daily scheduled rewards for the given date range', async () => {
+      await pgPools.stats.query(
+        'INSERT INTO daily_scheduled_rewards (day, participant_address, scheduled_rewards) VALUES ($1, $2, $3)',
+        ['2024-01-11', '0x20', '1']
+      )
+
+      const res = await fetch(
+        new URL(
+          '/participants/scheduled-rewards/daily?address=0x20&from=2024-01-11&to=2024-01-12',
```
bajtos, line 362: I only glanced at the new URL yesterday. When I looked more closely today, I found it a bit suboptimal. Since this endpoint is returning data about a single participant, I think it should have the participant address in the request path.

For example:

```
/participant/0x20/scheduled-rewards/daily?from=...&to=...
```

Notice singular "/participant", but I think plural "/participants" will work too.


On the second though, I think we won't need weekly/monthly rewards because there are not that many data points and it's trivial to aggregate daily rewards into monthly rewards - just sum the ~30 numbers.

In that light, I propose to rename the endpoint as follows:

```
/participant/{address}/daily-scheduled-rewards
```

or even

```
/participant/{address}/scheduled-rewards
```

(Notice that `GET /retrieval-success-rate` returns daily values but does not mention "daily" in the URL.)

Anyhow, this is out of scope of this pull request, let me know if you prefer me to re-post this comment elsewhere or open a new GH issue.

stats/test/handler.test.js
```diff
@@ -353,17 +350,40 @@ describe('HTTP request handler', () => {
...
+      assert.deepStrictEqual(stats, [
+        { day: '2024-01-11', scheduled_rewards: '1' }
+      ])
+    })
+  })
+
   describe('GET /miners/retrieval-success-rate/summary', () => {
     it('returns a summary of miners RSR for the given date range', async () => {
       // before the range
-      await givenRetrievalStats(pgPool, { day: '2024-01-10', minerId: 'f1one', total: 10, successful: 1 })
```
bajtos, line 359: Can we pass `pgPools.evaluate` as the first argument? For consistency with other places, e.g.

```js
 givenDailyParticipants(pgPools.evaluate
```

stats/test/handler.test.js
juliangruber: All sounds right 👍  I wanted to make the address a path segment first, but then thought it would be the convention in this repo to use query strings for everything. I don't exactly recollect where I got this impression from.

Laying the groundwork for this change here: https://github.com/filecoin-station/spark-stats/pull/149

juliangruber: Will follow up in a separate PR, as suggested by you

juliangruber: https://github.com/filecoin-station/spark-stats/pull/154

PR #149, opened 2024-06-14
stats: refactor path handling

juliangruber:
In preparation for https://github.com/filecoin-station/spark-stats/pull/147#discussion_r1638158012, this makes the path handling more flexible (each route can use its own parsing logic, path segments can be parametrized) while keeping it short using the utility functions

### DIFF
--- a/stats/lib/handler.js
+++ b/stats/lib/handler.js
@@ -84,6 +84,9 @@ const handler = async (req, res, pgPools) => {
     await respond(fetchMinersRSRSummary)
   } else if (await handlePlatformRoutes(req, res, pgPools)) {
     // no-op, request was handled by handlePlatformRoute
+  } else if (req.method === 'GET' && pathname === '/') {
+    // health check - required by Grafana datasources
+    res.end('OK')
   } else {
     notFound(res)
   }
--- a/stats/lib/handler.js
+++ b/stats/lib/handler.js
@@ -43,6 +43,16 @@ const enableCors = (req, res) => {
   }
 }
 
+const respondWithFetchFn = (pathname, searchParams, res, pgPools) => fetchFn => {
+  return getStatsWithFilterAndCaching(
+    pathname,
+    searchParams,
+    res,
+    pgPools,
+    fetchFn
+  )
+}
+
 /**
  * @param {import('node:http').IncomingMessage} req
  * @param {import('node:http').ServerResponse} res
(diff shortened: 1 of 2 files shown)


stats/lib/handler.js
```diff
@@ -43,6 +43,16 @@ const enableCors = (req, res) => {
   }
 }
 
+const respondWithFetchFn = (pathname, searchParams, res, pgPools) => fetchFn => {
```
juliangruber, line 46: alternative name suggestions:

- `createRespondWithFetchFn`
- `curryRespondWithFetchFn`

bajtos: I like `createRespondWithFetchFn` much more 👍🏻

stats/lib/handler.js
```diff
@@ -51,29 +61,24 @@ const enableCors = (req, res) => {
 const handler = async (req, res, pgPools) => {
   // Caveat! `new URL('//foo', 'http://127.0.0.1')` would produce "http://foo/" - not what we want!
   const { pathname, searchParams } = new URL(`http://127.0.0.1${req.url}`)
-  const segs = pathname.split('/').filter(Boolean)
```
bajtos, line 54: Please preserve this. It's handling edge cases like `//` in the path or a path ending with a trailing slash.

For example:

```js
const endpoint = '/' + pathname.split('/').filter(Boolean).join('/')
if (req.method === 'GET' && endpoint === '/deals/daily') {
// etc.
```

We can build on top of this a bit more, but I am not sure if that's not too much?

```js
const endpoint = req.method + ' /' + pathname.split('/').filter(Boolean).join('/')
if (endpoint === 'GET /deals/daily') {
```

bajtos (changes requested):
I love the general direction ❤️ 

Let's discuss the implementation details.

stats/lib/handler.js
juliangruber: No tests failed so I assumed this was fine. I remember having discussion about this, so I changed it back - albeit in a different way. What do you think?

juliangruber:
Going ahead here now as this is blocking other work (fix scheduled rewards URL, add participant reward transfers). Hindsight please, @bajtos 🙏

stats/lib/handler.js
bajtos: I like the new way 👍🏻

bajtos (commented):
LGTM 👍🏻

PR #150, opened 2024-06-15
observer: remove empty `index.js`

juliangruber:
This is convention that we have in other repos.

I stumbled over this blank `index.js` file a couple times.


bajtos (changes requested):
I have to say I find our current convention of having a small part of code in `/index.js` and everything else in `lib/*.js` files as confusing. I guess this convention works well for tiny packages, especially if there is only one `index.js` file. In our project, I never know whether I should look in `index.js` or into one of the `lib` files. This problem is amplified by the fact that `/lib/` and `/index.js` are usually very far away in the filesystem explorer in IDE, and `/index.js` is kind of lost among other root-level files.

Screenshot from spark-evaluate to illustrate what I mean:

Screenshot 2024-06-19 at 17 08 16

For a long time, I wanted to raise this point and agree on a better structure, but it never felt urgent enough. Until now, I guess 😄 

Let's find a solution that will work for both of us.

What do you think about creating a miniature top-level `index.js` file that re-exports the public API from different `lib/*.js` files?

juliangruber:
Ok let's discuss!

I think this PR makes it already better as before, as at least there's some content in index.js. We could also remove it instead, I'd be happy with that too.

What would you propose?

> What do you think about creating a miniature top-level index.js file that re-exports the public API from different lib/*.js files?

Are you talking about lib repos, or service repos like this too? What would be exported?

PatrickNercessian (commented):
Unless there's a hard need for an index.js file based on NodeJS or something that I'm unaware of, I'd lean towards removing the index.js, for the reasons @bajtos mentioned (mostly, in my expereince, in the IDE, index.js is often far away and at a different directory hierarchy level)

bajtos:
> I think this PR makes it already better as before, as at least there's some content in index.js. We could also remove it instead, I'd be happy with that too.
> 
> What would you propose?

Oh, I didn't realise we have an empty index.js file now. I prefer to remove it then.

Alternatively, in line with what I am writing below, we can put this code into `index.js`:

```js
export * from './lib/platform-stats.js'
```

Although, that feels like adding an unnecessary redirection layer since we don't expect anybody to import from this package. I think it's better to remove the index file and let the bin script import directly from lib files.

> > What do you think about creating a miniature top-level index.js file that re-exports the public API from different lib/*.js files?
> 
> Are you talking about lib repos, or service repos like this too? What would be exported?

I didn't think about this much. When the package needs more than one source code file, I prefer to put everything into `lib` so that all files are together.

I think such an approach works well for services, where "bin" scripts can import directly from various "lib" files as needed.

For a library, it makes sense to have a single index file exporting the package's public API. I like to keep this index file minimal; it should just re-export stuff from other lib files.

I would even consider placing such an index file into `lib/index.js`, but the top-level `index.js` location also works since the public API (and this file) should not be changing too often.

> Also, conceptually, it seems in a weird limbo state. bin files and lib files have nice conceptual narratives as to their function, and index.js not so much (most of it can semantically fit into the lib folder quite well I think, from what I am looking at our various services)

I love this framing - lib files implement the functionality, and bin files wrap that functionality into CLI/service executables. 👍🏻

bajtos (approved):
Nice and clean 😆 

Could you please update the pull request title to match the actual diff?

PR #152, opened 2024-06-15
Remove `EndablePgPools`

juliangruber:
We never use a `PgPools` that isn't endable

### DIFF
--- a/stats/test/handler.test.js
+++ b/stats/test/handler.test.js
@@ -26,7 +26,8 @@ describe('HTTP request handler', () => {
     const handler = createHandler({
       pgPools: {
         stats: null,
-        evaluate: pgPool
+        evaluate: pgPool,
+        end: async () => {}
       },
       logger: {
         info: debug,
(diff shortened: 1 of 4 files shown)


juliangruber:
hindsight please :)

stats/test/handler.test.js
```diff
@@ -26,7 +26,8 @@ describe('HTTP request handler', () => {
     const handler = createHandler({
       pgPools: {
         stats: null,
-        evaluate: pgPool
+        evaluate: pgPool,
+        end: async () => {}
```
bajtos, line 30: Yeah, I didn't want to add a dummy `end` implementation in the tests, so I introduced `EndablePgPools`.

I think in this place, the `end` method should call `pgPool.end()`, shouldn't it?

```diff
- end: async () => {}
+ end: () => pgPool.end()
```

bajtos (commented):
👍🏻

stats/test/handler.test.js
juliangruber: it doesn't make a difference, the tests still finish. But I could add it for completeness?

bajtos: Either way is fine with me.

PR #151, opened 2024-06-15
add get daily participant reward transfers

juliangruber:
- [x] Make address a path parameter
- [x] Add tests

PR #154, opened 2024-06-15
Update scheduled rewards url

juliangruber:
As discussed in https://github.com/filecoin-station/spark-stats/pull/147#discussion_r1638158012

### DIFF
--- a/stats/lib/handler.js
+++ b/stats/lib/handler.js
@@ -51,7 +51,7 @@ const enableCors = (req, res) => {
  * @param {URLSearchParams} searchParams
  * @param {import('node:http').ServerResponse} res
  * @param {PgPools} pgPools
- * @returns {(fetchFn: (pgPools: PgPools, filter: DateRangeFilter, pathVariables: object, ) => Promise<any>, pathParams?: object) => Promise<void>}
+ * @returns {(fetchFn: (pgPools: PgPools, filter: DateRangeFilter, pathVariables: object) => Promise<any>, pathParams?: object) => Promise<void>}
  */
 const createRespondWithFetchFn =
 (pathname, searchParams, res, pgPools) =>
--- a/stats/lib/handler.js
+++ b/stats/lib/handler.js
@@ -46,7 +46,6 @@ const enableCors = (req, res) => {
 }
 
 /**
- * @template {import('./typings.d.ts').DateRangeFilter} FilterType
  * @param {string} pathname
  * @param {URLSearchParams} searchParams
  * @param {import('node:http').ServerResponse} res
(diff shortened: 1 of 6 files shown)


juliangruber:
@bajtos hindsight please

stats/lib/stats-fetchers.js
```diff
@@ -136,18 +136,15 @@ export const fetchParticipantChangeRates = async (pgPools, filter) => {
 
 /**
  * @param {PgPools} pgPools
- * @param {import('./typings.js').DateRangeFilter & {address: string}} filter
+ * @param {import('./typings.js').DateRangeFilter} filter
+ * @param {string} address
  */
-export const fetchParticipantScheduledRewards = async (pgPools, filter) => {
+export const fetchParticipantScheduledRewards = async (pgPools, { from, to }, address) => {
```
bajtos, line 142: As I mentioned in https://github.com/filecoin-station/spark-stats/pull/151, generic data (shared by all impact evaluators) that isn't specific to Spark should be exposed via routes & fetchers implemented in `stats/lib/platform-routes.js` and `stats/lib/platform-stats-fetcher.js`.

Could you please move "participant scheduled rewards" there?

stats/lib/handler.js
```diff
@@ -62,29 +73,30 @@ export const sanitizePathname = pathname => `/${pathname.split('/').filter(Boole
...
 const handler = async (req, res, pgPools) => {
   // Caveat! `new URL('//foo', 'http://127.0.0.1')` would produce "http://foo/" - not what we want!
-  let { pathname, searchParams } = new URL(`http://127.0.0.1${req.url}`)
-  pathname = sanitizePathname(pathname)
+  const { pathname, searchParams } = new URL(`http://127.0.0.1${req.url}`)
+  const segs = pathname.split('/').filter(Boolean)
+  const url = `/${segs.join('/')}`
 
   enableCors(req, res)
   const respond = createRespondWithFetchFn(pathname, searchParams, res, pgPools)
```
bajtos, line 81: I think we should pass the normalised pathname (the value stored in `url`) to the `pathname` argument. That was the behaviour before your change, if I am reading the code correctly.

The same comment applies to `stats/lib/platform-routes.js`.

Thoughts?

bajtos (commented):
Looks great at the high level 👏🏻 

I have a few comments to discuss, see below.

stats/lib/stats-fetchers.js
juliangruber: This isn't a platform feature, as it exposes scheduled rewards only from Spark. For Voyager it would come from voyager-stats, which would read from the voyager impact evaluator

stats/lib/handler.js
juliangruber: I think you're right, tests didn't catch it though. Will follow up

juliangruber: https://github.com/filecoin-station/spark-stats/pull/180

PR #159, opened 2024-06-20
feat: add wallet count endpoint

PatrickNercessian:


observer/test/observer.test.js
```diff
@@ -1,5 +1,4 @@
 import assert from 'node:assert'
-import { beforeEach, describe, it } from 'mocha'
```
PatrickNercessian, line 2: These were unnecessary from before, as it turns out mocha variables are global. This is consistent with other tests we have

bajtos:
@PatrickNercessian I'd like to better understand the difference between the data served by the existing endpoint `GET /participants/daily` and this new endpoint.

Daily participants are updated every round based on the measurements submitted by each participant. A participant = a unique wallet address.
See https://github.com/filecoin-station/spark-evaluate/blob/1d6032035ee7bc59ffff010c3df2701ff188525f/lib/platform-stats.js#L12-L13

Screenshot from the Spark Public Dashboard:

Screenshot 2024-06-20 at 17 11 19

Your new endpoint is counting data in `daily_reward_transfers` table. Isn't this table populated only once a week when we release the rewards? (For the remaining 6 days, your endpoint will return 0 wallets, IIUC). Also, this table won't track participants (wallet addresses) who don't reach the 0.1 FIL threshold for payouts, right?

Let's double-check with @patrickwoodhead and @juliangruber  what value we want to show in the charts. Is it the number of people receiving rewards every week? If yes, then your solution is the correct one 👍🏻

stats/lib/platform-stats-fetchers.js
```diff
@@ -53,3 +57,17 @@ export const fetchDailyRewardTransfers = async (pgPool, filter) => {
...
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchDailyWalletCount = async (pgPool, filter) => {
+  return await getDailyDistinctCount({
+    pgPool,
+    table: 'daily_reward_transfers',
+    column: 'to_address',
+    filter,
+    asColumn: 'wallet_count'
```
bajtos, line 71: Since this reports the number of wallets that received rewards payout on a given day, I feel the names "daily wallet count" and "wallet_count" are a bit misleading.

I think more accurate names would be something like "daily recipients of rewards" and "recipient_count".

Thoughts?

stats/test/platform-routes.test.js
```diff
@@ -185,6 +185,38 @@ describe('Platform Routes HTTP request handler', () => {
       ])
     })
   })
+
+  describe('GET /wallets/daily', () => {
+    it('returns daily count of unique wallets for the given date range', async () => {
+      await givenDailyRewardTransferMetrics(pgPools.stats, '2024-01-10', [
+        { toAddress: 'to1', amount: 100, lastCheckedBlock: 1 }
```
bajtos, line 192: Can we make the field `lastCheckedBlock` optional and let the helper `givenDailyRewardTransferMetrics()` fill in a reasonable default value?

patrickwoodhead:
> @PatrickNercessian I'd like to better understand the difference between the data served by the existing endpoint `GET /participants/daily` and this new endpoint.
> 
> Daily participants are updated every round based on the measurements submitted by each participant. A participant = a unique wallet address. See https://github.com/filecoin-station/spark-evaluate/blob/1d6032035ee7bc59ffff010c3df2701ff188525f/lib/platform-stats.js#L12-L13
> 
> Screenshot from the Spark Public Dashboard:
> 
> Screenshot 2024-06-20 at 17 11 19
> Your new endpoint is counting data in `daily_reward_transfers` table. Isn't this table populated only once a week when we release the rewards? (For the remaining 6 days, your endpoint will return 0 wallets, IIUC). Also, this table won't track participants (wallet addresses) who don't reach the 0.1 FIL threshold for payouts, right?
> 
> Let's double-check with @patrickwoodhead and @juliangruber what value we want to show in the charts. Is it the number of people receiving rewards payout every week? If yes, then your solution is the correct one 👍🏻

I don't think we want to have a count of how many participants received rewards into their wallet due to the reasons @bajtos mentions above. Instead we want a count of the number of FIL addresses that have submitted at least one job result on a given day.

bajtos:
Based on the discussion above, I believe this pull request is no longer necessary. I am closing it as rejected. @PatrickNercessian, feel free to re-open it if you disagree.

PR #164, opened 2024-06-24
feat: add endpoint for participants with highest measurement count

PatrickNercessian:
Link: https://github.com/space-meridian/roadmap/issues/118

### DIFF
--- a/stats/bin/spark-stats.js
+++ b/stats/bin/spark-stats.js
@@ -2,6 +2,7 @@ import '../lib/instrument.js'
 import http from 'node:http'
 import { once } from 'node:events'
 import { createHandler } from '../lib/handler.js'
+import { updateTopMeasurementStations } from '../lib/platform-stats-fetchers.js'
 import { getPgPools } from '@filecoin-station/spark-stats-db'
 
 const {
@@ -11,6 +12,12 @@ const {
 } = process.env
 
 const pgPools = await getPgPools()
+
+// Refresh the leaderboard's materialized view every 12 hours
+setInterval(async () => {
+  await updateTopMeasurementStations(pgPools.evaluate)
+}, 1000 * 60 * 60 * 12)
+
 const logger = {
   error: console.error,
   info: console.info,
--- a/stats/bin/spark-stats.js
+++ b/stats/bin/spark-stats.js
@@ -2,7 +2,6 @@ import '../lib/instrument.js'
 import http from 'node:http'
 import { once } from 'node:events'
 import { createHandler } from '../lib/handler.js'
-import { updateTopMeasurementStations } from '../lib/platform-stats-fetchers.js'
 import { getPgPools } from '@filecoin-station/spark-stats-db'
 
 const {
@@ -12,11 +11,6 @@ const {
 } = process.env
 
 const pgPools = await getPgPools()
-
-// Refresh the leaderboard's materialized view every 12 hours
-setInterval(async () => {
-  await updateTopMeasurementStations(pgPools.evaluate)
-}, 1000 * 60 * 60 * 12)
 
 const logger = {
   error: console.error,
--- a/stats/bin/spark-stats.js
+++ b/stats/bin/spark-stats.js
@@ -11,7 +11,6 @@ const {
 } = process.env
 
 const pgPools = await getPgPools()
-
 const logger = {
   error: console.error,
   info: console.info,
--- a/stats/lib/platform-routes.js
+++ b/stats/lib/platform-routes.js
@@ -3,7 +3,7 @@ import {
   fetchDailyStationCount,
   fetchMonthlyStationCount,
   fetchDailyRewardTransfers,
-  fetchTopMeasurementParticipants,
+  fetchParticipantsWithTopMeasurements,
   fetchDailyStationAcceptedMeasurementCount
 } from './platform-stats-fetchers.js'
 
@@ -32,7 +32,7 @@ export const handlePlatformRoutes = async (req, res, pgPools) => {
   } else if (req.method === 'GET' && url === '/measurements/daily') {
     await respond(pgPools.evaluate, fetchDailyStationAcceptedMeasurementCount)
   } else if (req.method === 'GET' && url === '/participants/top-measurements') {
-    await respond(pgPools.evaluate, fetchTopMeasurementParticipants)
+    await respond(pgPools.evaluate, fetchParticipantsWithTopMeasurements)
   } else if (req.method === 'GET' && url === '/transfers/daily') {
     await respond(pgPools.stats, fetchDailyRewardTransfers)
   } else {
--- a/stats/lib/platform-routes.js
+++ b/stats/lib/platform-routes.js
@@ -31,7 +31,7 @@ export const handlePlatformRoutes = async (req, res, pgPools) => {
     await respond(pgPools.evaluate, fetchMonthlyStationCount)
   } else if (req.method === 'GET' && url === '/measurements/daily') {
     await respond(pgPools.evaluate, fetchDailyStationAcceptedMeasurementCount)
-  } else if (req.method === 'GET' && url === '/stations/top-measurements') {
+  } else if (req.method === 'GET' && url === '/participants/top-measurements') {
     await respond(pgPools.evaluate, fetchTopMeasurementStations)
   } else if (req.method === 'GET' && url === '/transfers/daily') {
     await respond(pgPools.stats, fetchDailyRewardTransfers)
--- a/stats/lib/platform-routes.js
+++ b/stats/lib/platform-routes.js
@@ -3,6 +3,7 @@ import {
   fetchDailyStationCount,
   fetchMonthlyStationCount,
   fetchDailyRewardTransfers,
+  fetchTopMeasurementStations,
   fetchDailyStationAcceptedMeasurementCount
 } from './platform-stats-fetchers.js'
 
@@ -30,6 +31,8 @@ export const handlePlatformRoutes = async (req, res, pgPools) => {
     await respond(pgPools.evaluate, fetchMonthlyStationCount)
   } else if (req.method === 'GET' && url === '/measurements/daily') {
     await respond(pgPools.evaluate, fetchDailyStationAcceptedMeasurementCount)
+  } else if (req.method === 'GET' && url === '/stations/top-measurements') {
+    await respond(pgPools.evaluate, fetchTopMeasurementStations)
   } else if (req.method === 'GET' && url === '/transfers/daily') {
     await respond(pgPools.stats, fetchDailyRewardTransfers)
   } else {
--- a/stats/lib/platform-routes.js
+++ b/stats/lib/platform-routes.js
@@ -3,7 +3,7 @@ import {
   fetchDailyStationCount,
   fetchMonthlyStationCount,
   fetchDailyRewardTransfers,
-  fetchTopMeasurementStations,
+  fetchTopMeasurementParticipants,
   fetchDailyStationAcceptedMeasurementCount
 } from './platform-stats-fetchers.js'
 
@@ -32,7 +32,7 @@ export const handlePlatformRoutes = async (req, res, pgPools) => {
   } else if (req.method === 'GET' && url === '/measurements/daily') {
     await respond(pgPools.evaluate, fetchDailyStationAcceptedMeasurementCount)
   } else if (req.method === 'GET' && url === '/participants/top-measurements') {
-    await respond(pgPools.evaluate, fetchTopMeasurementStations)
+    await respond(pgPools.evaluate, fetchTopMeasurementParticipants)
   } else if (req.method === 'GET' && url === '/transfers/daily') {
     await respond(pgPools.stats, fetchDailyRewardTransfers)
   } else {
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -43,6 +43,22 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter) 
   return rows
 }
 
+/**
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchTopMeasurementStations = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT station_id, SUM(accepted_measurement_count) as total_accepted_measurement_count
+    FROM daily_stations
+    WHERE day >= $1 AND day <= $2
+    GROUP BY station_id
+    ORDER BY total_accepted_measurement_count DESC
+    LIMIT 100
+  `, [filter.from, filter.to])
+  return rows
+}
+
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
   const { rows } = await pgPool.query(`
     SELECT day::TEXT, SUM(amount) as amount
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -53,10 +53,17 @@ export const fetchTopMeasurementStations = async (pgPool, filter) => {
   return (await pgPool.query('SELECT * FROM top_measurement_stations_mv')).rows
 }
 
+/**
+ * @param {Queryable} pgPool
+ */
 export const updateTopMeasurementStations = async (pgPool) => {
   await pgPool.query('REFRESH MATERIALIZED VIEW top_measurement_stations_mv')
 }
 
+/**
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
   const { rows } = await pgPool.query(`
     SELECT day::TEXT, SUM(amount) as amount
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -1,3 +1,4 @@
+import assert from 'http-assert'
 import { getDailyDistinctCount, getMonthlyDistinctCount } from './request-helpers.js'
 
 /** @typedef {import('@filecoin-station/spark-stats-db').Queryable} Queryable */
@@ -49,9 +50,7 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter) 
  */
 export const fetchTopMeasurementStations = async (pgPool, filter) => {
   // Block expensive multi-day queries
-  if (filter.to !== filter.from) {
-    throw new Error('Multi-day queries are not supported for this endpoint')
-  }
+  assert(filter.to === filter.from, 400, 'Multi-day queries are not supported for this endpoint')
 
   const { rows } = await pgPool.query(`
   SELECT
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -48,13 +48,21 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter) 
  * @param {import('./typings.js').DateRangeFilter} filter
  */
 export const fetchTopMeasurementStations = async (pgPool, filter) => {
+  // Block expensive multi-day queries
+  if (filter.to !== filter.from) {
+    throw new Error('Multi-day queries are not supported for this endpoint')
+  }
+
   const { rows } = await pgPool.query(`
-    SELECT station_id, SUM(accepted_measurement_count) as total_accepted_measurement_count
-    FROM daily_stations
-    WHERE day >= $1 AND day <= $2
-    GROUP BY station_id
-    ORDER BY total_accepted_measurement_count DESC
-    LIMIT 100
+  SELECT
+    participant_address,
+    COUNT(DISTINCT inet_group) AS inet_group_count,
+    COUNT(DISTINCT station_id) AS station_count,
+    SUM(accepted_measurement_count) AS accepted_measurement_count
+  FROM daily_stations
+  WHERE day >= $1 AND day <= $2
+  GROUP BY participant_address
+  ORDER BY accepted_measurement_count DESC;
   `, [filter.from, filter.to])
   return rows
 }
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -47,10 +47,10 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter) 
  * @param {Queryable} pgPool
  * @param {import('./typings.js').DateRangeFilter} filter
  */
-export const fetchTopMeasurementStations = async (pgPool, filter) => {
+export const fetchTopMeasurementParticipants = async (pgPool, filter) => {
   // Ignore the filter for this query
   // Get the top measurement stations from the Materialized View
-  return (await pgPool.query('SELECT * FROM top_measurement_stations_mv')).rows
+  return (await pgPool.query('SELECT * FROM top_measurement_participants_yesterday_mv')).rows
 }
 
 /**
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -1,4 +1,3 @@
-import assert from 'http-assert'
 import { getDailyDistinctCount, getMonthlyDistinctCount } from './request-helpers.js'
 
 /** @typedef {import('@filecoin-station/spark-stats-db').Queryable} Queryable */
@@ -49,21 +48,13 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter) 
  * @param {import('./typings.js').DateRangeFilter} filter
  */
 export const fetchTopMeasurementStations = async (pgPool, filter) => {
-  // Block expensive multi-day queries
-  assert(filter.to === filter.from, 400, 'Multi-day queries are not supported for this endpoint')
+  // Ignore the filter for this query
+  // Get the top measurement stations from the Materialized View
+  return (await pgPool.query('SELECT * FROM top_measurement_stations_mv')).rows
+}
 
-  const { rows } = await pgPool.query(`
-  SELECT
-    participant_address,
-    COUNT(DISTINCT inet_group) AS inet_group_count,
-    COUNT(DISTINCT station_id) AS station_count,
-    SUM(accepted_measurement_count) AS accepted_measurement_count
-  FROM daily_stations
-  WHERE day >= $1 AND day <= $2
-  GROUP BY participant_address
-  ORDER BY accepted_measurement_count DESC;
-  `, [filter.from, filter.to])
-  return rows
+export const updateTopMeasurementStations = async (pgPool) => {
+  await pgPool.query('REFRESH MATERIALIZED VIEW top_measurement_stations_mv')
 }
 
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -48,7 +48,7 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter) 
  * @param {Queryable} pgPool
  * @param {import('./typings.js').DateRangeFilter} filter
  */
-export const fetchTopMeasurementParticipants = async (pgPool, filter) => {
+export const fetchParticipantsWithTopMeasurements = async (pgPool, filter) => {
   assert(filter.to === filter.from, 400, 'Multi-day queries are not supported for this endpoint')
   const yesterdayUTC = new Date(Date.now() - 24 * 60 * 60 * 1000).toISOString().split('T')[0]
   assert(filter.to === yesterdayUTC, 400, 'filter.to must be set to yesterday, other values are not supported yet')
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -55,13 +55,6 @@ export const fetchTopMeasurementStations = async (pgPool, filter) => {
 
 /**
  * @param {Queryable} pgPool
- */
-export const updateTopMeasurementStations = async (pgPool) => {
-  await pgPool.query('REFRESH MATERIALIZED VIEW top_measurement_stations_mv')
-}
-
-/**
- * @param {Queryable} pgPool
  * @param {import('./typings.js').DateRangeFilter} filter
  */
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -1,3 +1,4 @@
+import assert from 'http-assert'
 import { getDailyDistinctCount, getMonthlyDistinctCount } from './request-helpers.js'
 
 /** @typedef {import('@filecoin-station/spark-stats-db').Queryable} Queryable */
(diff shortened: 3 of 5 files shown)


PatrickNercessian:
Have not added tests yet

This would go hand-in-hand with a change in spark-evaluate to add this Index: 

`CREATE INDEX stations_day_accepted_count ON daily_stations (day, accepted_count DESC);`

What do we think about this approach? Because we're using an aggregate function, the index will not be perfectly helpful. The query will still need to run the full summation and at least partial sorting, I think.

Alternatively, we could create a new table that just updates the total sum for each station_id. However, I lean towards thinking that would be overkill, considering we already cache queries for 24h, so we are rarely running this query.

stats/lib/platform-stats-fetchers.js
```diff
@@ -43,6 +43,22 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter)
...
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchTopMeasurementStations = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT station_id, SUM(accepted_measurement_count) as total_accepted_measurement_count
+    FROM daily_stations
+    WHERE day >= $1 AND day <= $2
+    GROUP BY station_id
+    ORDER BY total_accepted_measurement_count DESC
+    LIMIT 100
```
PatrickNercessian, line 57: Do we want a limit?

stats/lib/platform-routes.js
```diff
@@ -30,6 +31,8 @@ export const handlePlatformRoutes = async (req, res, pgPools) => {
     await respond(pgPools.evaluate, fetchMonthlyStationCount)
   } else if (req.method === 'GET' && url === '/measurements/daily') {
     await respond(pgPools.evaluate, fetchDailyStationAcceptedMeasurementCount)
+  } else if (req.method === 'GET' && url === '/stations/top-measurements') {
```
PatrickNercessian, line 34: Would we prefer a different endpoint?

stats/lib/platform-stats-fetchers.js
```diff
@@ -43,6 +43,22 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter)
...
+  `, [filter.from, filter.to])
```
PatrickNercessian, line 58: Do we want to use time ranges, or just do for the full table history in all cases.

stats/lib/platform-routes.js
bajtos: This one looks good to me 👍🏻

stats/lib/platform-stats-fetchers.js
```diff
@@ -43,6 +43,22 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter)
...
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchTopMeasurementStations = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT station_id, SUM(accepted_measurement_count) as total_accepted_measurement_count
+    FROM daily_stations
+    WHERE day >= $1 AND day <= $2
+    GROUP BY station_id
+    ORDER BY total_accepted_measurement_count DESC
```
bajtos, line 56: Do you know how expensive this query will be? Are there any indexes we can add to make it run faster?

PatrickNercessian: I mention this in my above comment: https://github.com/filecoin-station/spark-stats/pull/164#issuecomment-2186798998

It doesn't look like an index will be perfect, but supposedly still helpful.

stats/lib/platform-stats-fetchers.js
bajtos: Yes, please! There are 7497754 unique station ids in our prod DB now.

```
spark_evaluate=# SELECT COUNT(station_id) FROM daily_stations;
  count  
---------
 9230806
(1 row)

spark_evaluate=# SELECT COUNT(DISTINCT station_id) FROM daily_stations;
  count  
---------
 7497754
(1 row)
```

I think the hard-coded limit 100 is good enough for now. We can increase it in the future or make it configurable via the query string (e.g. `?limit=1000`).

stats/lib/platform-stats-fetchers.js
bajtos: Well, it is expensive.

```
# EXPLAIN ANALYZE SELECT station_id, SUM(accepted_measurement_count) as total_accepted_measurement_count
FROM daily_stations
WHERE day >= (now() - INTERVAL '7 day') AND day <= now()
GROUP BY station_id
ORDER BY total_accepted_measurement_count DESC
;
                                                                QUERY PLAN                                                                
------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=856016.50..860259.08 rows=1697033 width=97) (actual time=6236.102..6507.571 rows=1854497 loops=1)
   Sort Key: (sum(accepted_measurement_count)) DESC
   Sort Method: external merge  Disk: 206928kB
   ->  HashAggregate  (cost=495180.10..571699.85 rows=1697033 width=97) (actual time=4152.936..5149.416 rows=1854497 loops=1)
         Group Key: station_id
         Planned Partitions: 128  Batches: 129  Memory Usage: 4369kB  Disk Usage: 253808kB
         ->  Seq Scan on daily_stations  (cost=0.00..378860.25 rows=2032620 width=93) (actual time=83.001..2948.713 rows=2128699 loops=1)
               Filter: ((day <= now()) AND (day >= (now() - '7 days'::interval)))
               Rows Removed by Filter: 7102107
 Planning Time: 0.105 ms
 JIT:
   Functions: 12
   Options: Inlining true, Optimization true, Expressions true, Deforming true
   Timing: Generation 1.820 ms, Inlining 15.963 ms, Optimization 70.940 ms, Emission 41.194 ms, Total 129.916 ms
 Execution Time: 6655.125 ms
```

stats/lib/platform-stats-fetchers.js
```diff
@@ -43,6 +43,22 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter)
...
 
+/**
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchTopMeasurementStations = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT station_id, SUM(accepted_measurement_count) as total_accepted_measurement_count
+    FROM daily_stations
+    WHERE day >= $1 AND day <= $2
```
bajtos, line 54: What data do we need for the leaderboard? Do we need to aggregate historical data (e.g. over the last 7 days) or is it enough to always use the data from the previous day only? Or even today's data only?

The shorter the interval, the faster and cheaper the query will be.

bajtos (changes requested):

stats/lib/platform-stats-fetchers.js
bajtos: > I mention this in my above comment: [#164 (comment)](https://github.com/filecoin-station/spark-stats/pull/164#issuecomment-2186798998)
> 
> It doesn't look like an index will be perfect, but supposedly still helpful.

Ah, I missed your comment. Thank you for the link!

PatrickNercessian: I have tried this myself with some test data, and mine is much faster, and I'm not sure why... Roughly 850ms. And adding an index as I described above seems to have no effect

```
import http from 'node:http'
import { once } from 'node:events'
import assert from 'node:assert'
import createDebug from 'debug'
import { getPgPools } from '@filecoin-station/spark-stats-db'

import { assertResponseStatus, getPort } from '../test/test-helpers.js'
import { createHandler } from '../lib/handler.js'

const debug = createDebug('test')

const givenDailyStationMetrics = async (pgPoolEvaluate, day, stationStats) => {
    await pgPoolEvaluate.query(`
      INSERT INTO daily_stations (day, station_id, accepted_measurement_count)
      SELECT $1 AS day, UNNEST($2::text[]) AS station_id, UNNEST($3::int[]) AS accepted_measurement_count
      ON CONFLICT DO NOTHING
      `, [
      day,
      stationStats.map(s => s.stationId),
      stationStats.map(s => s.acceptedMeasurementCount)
    ])
  }

let pgPools = await getPgPools()
await pgPools.evaluate.query('DELETE FROM daily_stations')

const handler = createHandler({
pgPools,
logger: {
    info: debug,
    error: console.error,
    request: debug
}
})

let server = http.createServer(handler)
server.listen()
await once(server, 'listening')
const baseUrl = `http://127.0.0.1:${getPort(server)}`

for (let day = '2024-01-01'; day <= '2024-01-09'; day = day.substring(0, 8) + (parseInt(day.substring(8)) + 1).toString().padStart(2, '0')) {
  console.log(day)
  const station_stats = []
  for (let i = 0; i < 200000; i++) {
    // await givenDailyStationMetrics(pgPools.evaluate, day, [
      station_stats.push({ stationId: `station${i}`, acceptedMeasurementCount: Math.floor(Math.random() * 1000) })
    // ])
  }
  await givenDailyStationMetrics(pgPools.evaluate, day, station_stats)
}

// assert 1,800,000 records were inserted
const { rows: [{ count }] } = await pgPools.evaluate.query('SELECT COUNT(*) FROM daily_stations')
assert.strictEqual(count, '1800000')

const startTime = Date.now()

const res = await fetch(
  new URL(
    '/stations/top-measurements?from=2024-01-01&to=2024-01-02',
    baseUrl
  ), {
    redirect: 'manual'
  }
)

console.log(await res.json())

const executionTime = Date.now() - startTime
console.log('Execution time:', executionTime)
```

stats/lib/platform-stats-fetchers.js
PatrickNercessian: @patrickwoodhead Any opinion here?

stats/lib/platform-stats-fetchers.js
PatrickNercessian: Ah looks like one difference is your query did not have a LIMIT. But even if I remove it, mine is only 2300ms

PatrickNercessian:
```
spark_evaluate=#     EXPLAIN ANALYZE SELECT station_id, SUM(accepted_measurement_count) as total_accepted_measurement_count
    FROM daily_stations
    WHERE day >= '2024-01-01' AND day <= '2024-01-09'
    GROUP BY station_id
    ORDER BY total_accepted_measurement_count DESC;
                                                                           QUERY PLAN                                                                            
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=210141.11..210639.40 rows=199316 width=21) (actual time=903.301..927.241 rows=200000 loops=1)
   Sort Key: (sum(accepted_measurement_count)) DESC
   Sort Method: external merge  Disk: 6672kB
   ->  Finalize GroupAggregate  (cost=138011.99..188508.62 rows=199316 width=21) (actual time=604.484..842.231 rows=200000 loops=1)
         Group Key: station_id
         ->  Gather Merge  (cost=138011.99..184522.30 rows=398632 width=21) (actual time=604.475..744.019 rows=579498 loops=1)
               Workers Planned: 2
               Workers Launched: 2
               ->  Sort  (cost=137011.96..137510.25 rows=199316 width=21) (actual time=595.138..625.003 rows=193166 loops=3)
                     Sort Key: station_id
                     Sort Method: external merge  Disk: 6512kB
                     Worker 0:  Sort Method: external merge  Disk: 6464kB
                     Worker 1:  Sort Method: external merge  Disk: 6344kB
                     ->  Partial HashAggregate  (cost=104597.25..115379.47 rows=199316 width=21) (actual time=334.234..453.020 rows=193166 loops=3)
                           Group Key: station_id
                           Planned Partitions: 4  Batches: 5  Memory Usage: 8241kB  Disk Usage: 15776kB
                           Worker 0:  Batches: 5  Memory Usage: 8241kB  Disk Usage: 15376kB
                           Worker 1:  Batches: 5  Memory Usage: 8241kB  Disk Usage: 15704kB
                           ->  Parallel Seq Scan on daily_stations  (cost=0.00..50691.00 rows=750000 width=17) (actual time=11.587..133.580 rows=600000 loops=3)
                                 Filter: ((day >= '2024-01-01'::date) AND (day <= '2024-01-09'::date))
 Planning Time: 0.225 ms
 Execution Time: 945.630 ms
(22 rows)

```

PatrickNercessian:
Depends on filecoin-station/spark-evaluate#274

stats/lib/platform-stats-fetchers.js
bajtos: I believe we agreed to show data for one day only, i.e. require that `filter.from === filter.to`.

PatrickNercessian: This was an old question, irrelevant now! Sorry, should have marked it resolved

stats/lib/platform-stats-fetchers.js
```diff
@@ -43,6 +44,28 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter)
...
+  const { rows } = await pgPool.query(`
+  SELECT
+    participant_address,
+    COUNT(DISTINCT inet_group) AS inet_group_count,
+    COUNT(DISTINCT station_id) AS station_count,
+    SUM(accepted_measurement_count) AS accepted_measurement_count
+  FROM daily_stations
+  WHERE day >= $1 AND day <= $2
+  GROUP BY participant_address
+  ORDER BY accepted_measurement_count DESC;
```
bajtos, line 64: Here is the output of this query for the partial data for today, after landing https://github.com/filecoin-station/spark-evaluate/pull/274.

```
            participant_address             | inet_group_count | station_count | accepted_measurement_count 
--------------------------------------------+------------------+---------------+----------------------------
 pre_column_addition                        |                1 |        154159 |                    1036555
 0xb63153ae08c1d7b4f0dee0b0df725f3a9b8cdaae |              274 |           292 |                       6414
 0x170bCcAe27f8160dDE0CbeF147BDc408d9430e69 |              398 |         33062 |                       3995
 0xc04C7538632a03f7F8Ab651C2363B0641f934612 |              398 |         33066 |                       3877
 0xBA2ff3dA97dF5CA0f8Aa3Fa49F5AbcbD4A5eD36C |              398 |         33036 |                       3832
 0x0aeDc88f2f3aebAE54b9b191E323eEEC0B76c1AF |              114 |           123 |                       2464
 0x7581D288f51622901eaB1d340be773241c53D025 |              129 |           131 |                       2455
 0xCE53e6dEfbD1f0810BFcd691660ad311d0DB24f7 |              107 |           120 |                       2153
 0xB7bAa0811340d2Fc12c7C08eC016993B92C368a9 |               92 |            92 |                       1558
(...)
```

Notice that we treat `inet_group` value `pre_column_addition` as a valid subnet.

And the top participant is the famous `pre_column_addition` 😆 

This won't be a problem from tomorrow & onwards as long as we never look back at the data for today (e.g. if we decide to build data for the last 30 days).

I propose to delete all rows with `pre_column_addition`. We will lose some historical data, but that's not a big deal IMO, as we don't have the full history anyways. In hindsight, we should have deleted those columns as part of the DB schema migration 🙈 

WDYT, @PatrickNercessian @juliangruber?

stats/lib/platform-stats-fetchers.js
bajtos: We agreed to use the data from the previous day only. Marking this thread as resolved.

stats/lib/platform-stats-fetchers.js
bajtos: Performance analysis of the new query - `Execution Time: 1924.542 ms`

```
spark_evaluate=# EXPLAIN ANALYZE SELECT
    participant_address,
    COUNT(DISTINCT inet_group) AS inet_group_count,
    COUNT(DISTINCT station_id) AS station_count,
    SUM(accepted_measurement_count) AS accepted_measurement_count
  FROM daily_stations
  WHERE day >= DATE(now()) AND day <= DATE(now())
  GROUP BY participant_address
  ORDER BY accepted_measurement_count DESC;
                                                                               QUERY PLAN                                           
                                    
------------------------------------------------------------------------------------------------------------------------------------
------------------------------------
 Sort  (cost=103736.43..103736.93 rows=200 width=56) (actual time=1912.549..1912.978 rows=3071 loops=1)
   Sort Key: (sum(accepted_measurement_count)) DESC
   Sort Method: quicksort  Memory: 528kB
   ->  GroupAggregate  (cost=101834.29..103728.79 rows=200 width=56) (actual time=963.668..1911.692 rows=3071 loops=1)
         Group Key: participant_address
         ->  Sort  (cost=101834.29..102212.79 rows=151400 width=157) (actual time=963.616..1024.363 rows=333525 loops=1)
               Sort Key: participant_address
               Sort Method: external merge  Disk: 50496kB
               ->  Index Scan using daily_stations_pkey on daily_stations  (cost=0.69..81835.78 rows=151400 width=157) (actual time=
7.956..723.050 rows=333525 loops=1)
                     Index Cond: ((day >= date(now())) AND (day <= date(now())))
 Planning Time: 0.156 ms
 JIT:
   Functions: 11
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 1.683 ms, Inlining 0.000 ms, Optimization 0.447 ms, Emission 7.201 ms, Total 9.331 ms
 Execution Time: 1924.542 ms
(16 rows)
```

bajtos: Experiment: 

```sql
CREATE INDEX mbtest ON daily_stations (day, participant_address);
```

Performance - `Execution Time: 1536.079 ms` - a very small improvement.

```
 Sort  (cost=96047.44..96047.94 rows=200 width=56) (actual time=1526.023..1526.609 rows=3071 loops=1)
   Sort Key: (sum(accepted_measurement_count)) DESC
   Sort Method: quicksort  Memory: 528kB
   ->  GroupAggregate  (cost=94153.83..96039.79 rows=200 width=56) (actual time=361.709..1525.309 rows=3071 loops=1)
         Group Key: participant_address
         ->  Sort  (cost=94153.83..94530.62 rows=150717 width=157) (actual time=361.682..446.334 rows=333525 loops=1)
               Sort Key: participant_address
               Sort Method: external merge  Disk: 50512kB
               ->  Index Scan using mbtest on daily_stations  (cost=0.45..74249.75 rows=150717 width=157) (actual time=0.038..176.38
4 rows=333525 loops=1)
                     Index Cond: ((day >= date(now())) AND (day <= date(now())))
 Planning Time: 0.219 ms
 Execution Time: 1536.079 ms
```

bajtos: Experiment 2:

```sql
```

Performance - `Execution Time: 2416.290 ms` 👎🏻 

```
 Sort  (cost=104484.10..104484.60 rows=200 width=56) (actual time=2403.051..2403.494 rows=3093 loops=1)
   Sort Key: (sum(accepted_measurement_count)) DESC
   Sort Method: quicksort  Memory: 531kB
   ->  GroupAggregate  (cost=102570.31..104476.46 rows=200 width=56) (actual time=1273.335..2402.412 rows=3093 loops=1)
         Group Key: participant_address
         ->  Sort  (cost=102570.31..102951.14 rows=152332 width=157) (actual time=1273.285..1352.646 rows=447542 loops=1)
               Sort Key: participant_address
               Sort Method: external merge  Disk: 68584kB
               ->  Index Scan using daily_stations_pkey on daily_stations  (cost=0.69..82441.81 rows=152332 width=157) (actual time=
6.874..966.039 rows=447542 loops=1)
                     Index Cond: ((day >= date(now())) AND (day <= date(now())))
 Planning Time: 0.095 ms
 JIT:
   Functions: 11
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 1.355 ms, Inlining 0.000 ms, Optimization 0.332 ms, Emission 6.294 ms, Total 7.981 ms
 Execution Time: 2416.290 ms
```

bajtos: Measuring the performance with no index again - `Execution Time: 2546.602 ms`.

I think as we are recording more data, the query becomes even more expensive when there is no index.

```
 Sort  (cost=104484.10..104484.60 rows=200 width=56) (actual time=2531.676..2532.188 rows=3093 loops=1)
   Sort Key: (sum(accepted_measurement_count)) DESC
   Sort Method: quicksort  Memory: 531kB
   ->  GroupAggregate  (cost=102570.31..104476.46 rows=200 width=56) (actual time=1366.055..2531.003 rows=3093 loops=1)
         Group Key: participant_address
         ->  Sort  (cost=102570.31..102951.14 rows=152332 width=157) (actual time=1366.008..1446.075 rows=447542 loops=1)
               Sort Key: participant_address
               Sort Method: external merge  Disk: 68584kB
               ->  Index Scan using daily_stations_pkey on daily_stations  (cost=0.69..82441.81 rows=152332 width=157) (actual time=
7.966..1031.773 rows=447542 loops=1)
                     Index Cond: ((day >= date(now())) AND (day <= date(now())))
 Planning Time: 0.155 ms
 JIT:
   Functions: 11
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 1.583 ms, Inlining 0.000 ms, Optimization 0.425 ms, Emission 7.309 ms, Total 9.317 ms
 Execution Time: 2546.602 ms
```

bajtos: Let's see how long this query will take on Monday.

I suspect we will need to add an index `ON daily_stations (day, participant_address)`

@PatrickNercessian What other options do we have to make this query run faster? You mentioned materialized views, how would that work?

PatrickNercessian: This will not necessarily make it faster, but moreso invisible to the any user. The real issue with any slow query today is that the first cache miss will take a long time.

A Materialized View (as I understand it), is basically a virtual table that is constructed by a normal query on other tables on a schedule. Then our actual API queries are basically instant, since they're just reading from the table (e.g. `SELECT * FROM top_measurement_stations_mv`)

For example,

```
CREATE MATERIALIZED VIEW top_measurement_stations_mv AS
SELECT
  participant_address,
  COUNT(DISTINCT inet_group) AS inet_group_count,
  COUNT(DISTINCT station_id) AS station_count,
  SUM(accepted_measurement_count) AS accepted_measurement_count
FROM daily_stations
WHERE day = CURRENT_DATE - INTERVAL '1 day'
GROUP BY participant_address
ORDER BY accepted_measurement_count DESC;
```

This may not be perfect, but the idea above works. Then we can run 

`REFRESH MATERIALIZED VIEW top_measurement_stations_mv;`

on a schedule, such as once per day.

stats/lib/platform-stats-fetchers.js
PatrickNercessian: I am always apprehensive toward deleting data, (not that it will break anything, but moreso "maybe we'll want it later!") but that's not a big deal.

An alternative could be something like:

```
  SELECT
    participant_address,
    COUNT(DISTINCT CASE WHEN inet_group != 'pre_column_addition' THEN inet_group END) AS inet_group_count,
    COUNT(DISTINCT station_id) AS station_count,
    SUM(accepted_measurement_count) AS accepted_measurement_count
  FROM daily_stations
  WHERE day >= $1 AND day <= $2
    AND participant_address != 'pre_column_addition'
    AND inet_group != 'pre_column_addition'
  GROUP BY participant_address
  ORDER BY accepted_measurement_count DESC;
```

But again, no hard preference from me.

stats/lib/platform-stats-fetchers.js
bajtos: I am running the query now to fetch stats for yesterday.

```
spark_evaluate=# EXPLAIN ANALYZE SELECT
  participant_address,
  COUNT(DISTINCT inet_group) AS inet_group_count,
  COUNT(DISTINCT station_id) AS station_count,
  SUM(accepted_measurement_count) AS accepted_measurement_count
FROM daily_stations
WHERE day >= DATE(now() - INTERVAL '1 day') AND day <= DATE(now() - INTERVAL '1 day')
GROUP BY participant_address
ORDER BY accepted_measurement_count DESC;
                                                                            QUERY PLAN                                                                            
------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=2311999.04..2311999.12 rows=32 width=59) (actual time=33629.984..33689.545 rows=3408 loops=1)
   Sort Key: (sum(accepted_measurement_count)) DESC
   Sort Method: quicksort  Memory: 576kB
   ->  GroupAggregate  (cost=1512818.68..2311998.24 rows=32 width=59) (actual time=6176.119..33688.200 rows=3408 loops=1)
         Group Key: participant_address
         ->  Gather Merge  (cost=1512818.68..2248804.95 rows=6319297 width=144) (actual time=6176.076..8616.084 rows=6274167 loops=1)
               Workers Planned: 2
               Workers Launched: 2
               ->  Sort  (cost=1511818.66..1518401.26 rows=2633040 width=144) (actual time=6064.667..6481.303 rows=2091389 loops=3)
                     Sort Key: participant_address
                     Sort Method: external merge  Disk: 329000kB
                     Worker 0:  Sort Method: external merge  Disk: 326776kB
                     Worker 1:  Sort Method: external merge  Disk: 338912kB
                     ->  Parallel Seq Scan on daily_stations  (cost=0.00..1009635.55 rows=2633040 width=144) (actual time=197.807..4200.450 rows=2091389 loops=3)
                           Filter: ((day >= date((now() - '1 day'::interval))) AND (day <= date((now() - '1 day'::interval))))
                           Rows Removed by Filter: 8065788
 Planning Time: 0.181 ms
 JIT:
   Functions: 15
   Options: Inlining true, Optimization true, Expressions true, Deforming true
   Timing: Generation 3.134 ms, Inlining 173.461 ms, Optimization 171.399 ms, Emission 118.043 ms, Total 466.037 ms
 Execution Time: 33731.846 ms
(22 rows)
```

bajtos: With index `ON daily_stations (day, participant_address)`:

```
 Planning Time: 0.222 ms
 JIT:
   Functions: 15
   Options: Inlining true, Optimization true, Expressions true, Deforming true
   Timing: Generation 3.101 ms, Inlining 161.029 ms, Optimization 168.619 ms, Emission 117.287 ms, Total 450.036 ms
 Execution Time: 38431.564 ms
```

Notice that Postgres is not able to use an index for the initial filtering by date and always runs a Parallel Seq Scan 😞

bajtos: FWIW, there were 3,408 active participants yesterday.

As I understand the output of `EXPLAIN ANALYZE`, the most expensive operation is sorting the results according to `accepted_measurement_count` - it takes about 30 seconds from the total query time of 36 seconds.

stats/lib/platform-stats-fetchers.js
bajtos: I decided to delete the data with `pre_column_addition`. The SQL query is already too expensive and difficult to optimise via index; let's not make it even more complex & slower.

stats/lib/platform-stats-fetchers.js
bajtos: There were 3,408 active participants yesterday. I think returning ~3.5k items is acceptable.

stats/lib/platform-stats-fetchers.js
```diff
@@ -43,6 +43,20 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter)
...
+export const fetchTopMeasurementStations = async (pgPool, filter) => {
+  // Ignore the filter for this query
+  // Get the top measurement stations from the Materialized View
+  return (await pgPool.query('SELECT * FROM top_measurement_stations_mv')).rows
+}
+
+export const updateTopMeasurementStations = async (pgPool) => {
+  await pgPool.query('REFRESH MATERIALIZED VIEW top_measurement_stations_mv')
+}
+
```
PatrickNercessian, line 59: In order to have this reused in both tests and spark-stats.js, I made it a function even though it's one simple line.

I am not sure this is the best place for it, but no where else seemed great. Let me know if you have any thoughts

stats/bin/spark-stats.js
```diff
@@ -11,6 +12,12 @@ const {
 } = process.env
 
 const pgPools = await getPgPools()
+
+// Refresh the leaderboard's materialized view every 12 hours
+setInterval(async () => {
+  await updateTopMeasurementStations(pgPools.evaluate)
+}, 1000 * 60 * 60 * 12)
```
bajtos, line 19: IIRC, spark-stats has read-only access to the DB; it's connected to a reader replica to reduce the load on the main DB server instance.

PatrickNercessian: Moved it to spark-evaluate PR!

stats/lib/platform-stats-fetchers.js
juliangruber: Agreed, a 30s query is really heavy on the backend and creates a bad experience when uncached

stats/test/platform-routes.test.js
```diff
@@ -153,6 +157,50 @@ describe('Platform Routes HTTP request handler', () => {
...
+      await givenDailyStationMetrics(pgPools.evaluate, today.toISOString().split('T')[0], [
+        { ...STATION_STATS, acceptedMeasurementCount: 10 }
+      ])
+
+      await pgPools.evaluate.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
+
+      // We don't care about the date range for this query
+      const res = await fetch(
+        new URL(
+          '/participants/top-measurements?from=2024-01-11&to=2024-01-11',
```
bajtos, line 181: > We don't care about the date range for this query

As I am arguing in my other comment above, we need to care about the date range.

bajtos (changes requested):
You are going in the right direction 👍🏻

stats/lib/platform-routes.js
```diff
@@ -3,6 +3,7 @@ import {
   fetchDailyStationCount,
   fetchMonthlyStationCount,
   fetchDailyRewardTransfers,
+  fetchTopMeasurementParticipants,
```
juliangruber, line 6: topMeasurementParticipants is a bit awkward, what about topMeasurementsParticipants?

juliangruber: or fetchParticipantsWithTopMeasurements

stats/lib/platform-stats-fetchers.js
```diff
@@ -43,6 +43,20 @@ export const fetchDailyStationAcceptedMeasurementCount = async (pgPool, filter)
   return rows
 }
 
+/**
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchTopMeasurementParticipants = async (pgPool, filter) => {
+  // Ignore the filter for this query
```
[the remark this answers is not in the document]
juliangruber: > When I request top participants using a date range for today, the API will return the response using data from yesterday. 

I think this can't happen because if filter=today then the assert above will throw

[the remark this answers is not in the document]
juliangruber: ah I see this comment is old. Will resolve

juliangruber (approved):

stats/lib/platform-routes.js
PatrickNercessian: Sure, addressed in https://github.com/filecoin-station/spark-stats/pull/164/commits/cf96533b9c84f7a2e4b9cfe2dc871bd6be1c1831

bajtos (approved):
I realised that since we don't use the filter at all, maybe this endpoint shouldn't be implemented via `respond(pgPools.evaluate, fetchParticipantsWithTopMeasurements)`. 

Instead, we could write a custom request handler and configure a caching strategy that makes more sense in this particular case - e.g., we can set TTL until midnight UTC.

Anyhow, it's a relatively minor thing, I am fine to land this pull request as it is now.

:shipit:

bajtos:
The CI build is failing in the dry-run step because this PR does not have access to GLIF_TOKEN secret.

```
Error: server response 401 Unauthorized (
request={  }, 
response={  }, 
error=null, 
info={ "requestUrl": "https://api.node.glif.io/rpc/v0", "responseBody": "<html>\r\n<head><title>401 Authorization Required</title></head>\r\n<body>\r\n<center><h1>401 Authorization Required</h1></center>\r\n</body>\r\n</html>\r\n", "responseStatus": "401 Unauthorized" }, 
code=SERVER_ERROR, 
version=6.13.1
)
```

PR #180, opened 2024-07-16
fix pass normalized pathname to handler

juliangruber:
Ref https://github.com/filecoin-station/spark-stats/pull/154#discussion_r1679087445


juliangruber:
@bajtos please follow up with tests if you think this feature is important enough. For me it's a nice to have but not necessary, so atm I don't see it as a good use of my time

bajtos (approved):

PR #185, opened 2024-07-23
 Types: Ensure correct db is passed

juliangruber:


db/index.js
```diff
@@ -6,6 +6,8 @@ import Postgrator from 'postgrator'
 
 // re-export types
 /** @typedef {import('./typings.js').PgPools} PgPools */
+/** @typedef {import('./typings.js').PgPoolStats} PgPoolStats */
+/** @typedef {import('./typings.js').PgPoolEvaluate} PgPoolEvaluate */
 /** @typedef {import('./typings.js').Queryable} Queryable */
```
bajtos, line 11: FWIW, with the latest TypeScript version, we should be able to simplify this block as follows:

```js
/** @import { PgPools, PgPoolStats, PgPoolEvaluate, Queryable } from './typings.js' */
```

See https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#the-jsdoc-import-tag

bajtos: Nevermind. Such an import statement does not re-export the types.

bajtos:
@juliangruber PTAL.

I have two versions to pick from - the first one uses `getNamedPool` helper (up to the commit [c2fb0d3](https://github.com/filecoin-station/spark-stats/pull/185/commits/c2fb0d3b3e53c3be8a0be015ea20b6e6ef7cacd6)), the second one calls Object.assign directly (the final version). I am fine with both of them, which one do you like more?

juliangruber:
Awesome thank you! I like the one without the helper more, since its logic is so simple

bajtos (approved):
:shipit:

PR #210, opened 2024-09-04
feat: ingest platform stats from new spark-evaluate schema

PatrickNercessian:
Links:
* https://github.com/filecoin-station/spark-evaluate/pull/311
* https://github.com/space-meridian/roadmap/issues/132


observer/test/observer.test.js
```diff
@@ -7,8 +7,7 @@ import { observeTransferEvents, observeScheduledRewards } from '../lib/observer.
 
 describe('observer', () => {
   let pgPools
-  const getDayAsISOString = d => d.toISOString().split('T')[0]
-  const today = () => getDayAsISOString(new Date())
+  const today = () => new Date().toLocaleDateString('en-CA')
```
PatrickNercessian, line 10: The existing way was causing a mismatch between the hours of 8pm - 12am in EST, because it would use UTC, while the SQL now() calls default to the local time zone.

What do you guys think of this change here, and the similar one in request-helpers.js?

stats/lib/platform-stats-fetchers.js
```diff
@@ -13,25 +8,27 @@ import {
...
+  `, [filter.from, filter.to])
+  return rows
 }
 
 /**
  * @param {Queryable} pgPool
  * @param {import('./typings.js').DateRangeFilter} filter
  */
 export const fetchMonthlyStationCount = async (pgPool, filter) => {
-  return await getMonthlyDistinctCount({
```
PatrickNercessian, line 29: These functions are now only used once each: https://github.com/filecoin-station/spark-stats/blob/a6455da4c58810a3e32b83adbaaef7dd5ba8ecdd/stats/lib/stats-fetchers.js#L67-L85

Would you prefer to remove these function definitions and place the logic directly in their calling function?

PatrickNercessian:
The current error in build-stats step seems to be that there is no `day` column in the materialized view, which seems like an issue with using an older version of spark-evaluate. However, #207 was supposed to handle this. Any ideas? @bajtos

bajtos:
> The current error in build-stats step seems to be that there is no `day` column in the materialized view, which seems like an issue with using an older version of spark-evaluate. However, #207 was supposed to handle this. Any ideas? @bajtos

IIRC, we run the "refresh" queries every 12 hours. Maybe you hit the window when the change was deployed, but the materialised view has not been updated yet.

When I query the materialized view now, I see that it includes the `day` column.

Let me re-run the CI.

bajtos:
> > When I query the materialized view now, I see that it includes the `day` column.
> 
> I woke up this morning realizing that the PR that mattered for this error was actually [filecoin-station/spark-evaluate#341](https://github.com/filecoin-station/spark-evaluate/pull/341), which was just now merged, so the error was expected actually. My bad!

Makes sense 👍🏻 

I misunderstood your original comment. Now I understand this issue is the local database used by `npm test`, which makes sense.

Can you update spark-evaluate version as part of this PR (similarly to what I did in #207)?

observer/test/observer.test.js
bajtos: IIUC, you are changing today from "today in UTC" to "today in Canada". Isn't this going to break the tests for people in Europe/Asia/Australia?

> the SQL now() calls default to the local time zone

That depends on the time zone configured on the machine running the server. I am using Docker and `now()` returns time in UTC - the same value as `new Date().toISOString()` in JavaScript.

Anyhow, I support fixing the way how we build test data to work in all time zones.

Can we make a DB query to give us the current date?

Something along the following lines:

```js
let today

// ...
before(async() => {
  // ...
  today = (await pgPools.stats.query('SELECT CURRENT_DATE AS today')).rows[0].today
})
```

stats/lib/platform-stats-fetchers.js
```diff
@@ -13,25 +8,27 @@ import {
...
 export const fetchDailyStationCount = async (pgPool, filter) => {
-  return await getDailyDistinctCount({
-    pgPool,
-    table: 'daily_stations',
-    column: 'station_id',
-    filter
-  })
+  const { rows } = await pgPool.query(`
+    SELECT day::TEXT, station_count
+    FROM daily_measurements_summary
```
bajtos, line 13: I was confused when I saw this first time. I wish we used a more accurate name than `daily_measurements_summary`.

Changing this table name after we land this PR will be more difficult, because we need to update both spark-evaluate and spark-stats repo in tandem.

@juliangruber @PatrickNercessian 
What do you think about renaming that table first, before we land this PR? (Assuming you agree we should find a better name.) 

Proposed name: `daily_summary`. Alternatively: `daily_network_stats`

stats/lib/platform-stats-fetchers.js
bajtos: It's great that you pointed this out, I would not realise that these helpers are used in a single place.

I am fine either way.

Since we already created these helpers, there is little (if any) maintenance overhead, and we may have a use for them in the future, I slightly prefer to keep them around.

@juliangruber thoughts?

stats/lib/platform-stats-fetchers.js
```diff
@@ -60,7 +54,9 @@ export const fetchParticipantsWithTopMeasurements = async (pgPool, filter) => {
   assert(filter.to === yesterday(), 400, 'filter.to must be set to yesterday, other values are not supported yet')
   // Ignore the filter for this query
   // Get the top measurement stations from the Materialized View
-  return (await pgPool.query('SELECT * FROM top_measurement_participants_yesterday_mv')).rows
+  return (await pgPool.query(`
+    SELECT day::TEXT, participant_address, station_count, accepted_measurement_count, inet_group_count
+    FROM top_measurement_participants_yesterday_mv`)).rows
```
bajtos, line 59:
```suggestion
  return (await pgPool.query(`
    SELECT day::TEXT, participant_address, station_count, accepted_measurement_count, inet_group_count
    FROM top_measurement_participants_yesterday_mv
  `)).rows
```

stats/lib/request-helpers.js
```diff
@@ -5,10 +5,10 @@ import pg from 'pg'
 
 /** @typedef {import('@filecoin-station/spark-stats-db').Queryable} Queryable */
 
-export const getDayAsISOString = (d) => d.toISOString().split('T')[0]
+export const getDateString = (d) => d.toLocaleDateString('en-CA')
 
-export const today = () => getDayAsISOString(new Date())
-export const yesterday = () => getDayAsISOString(new Date(Date.now() - 24 * 60 * 60 * 1000))
+export const today = () => getDateString(new Date())
+export const yesterday = () => getDateString(new Date(Date.now() - 24 * 60 * 60 * 1000))
```
bajtos, line 11: This may need a change depending on the outcome of https://github.com/filecoin-station/spark-stats/pull/210/files#r1742924839

observer/test/observer.test.js
bajtos: I think I don't fully understand what exactly is performed by `new Date().toLocaleDateString('en-CA')`.

Are you using CA instead of e.g. US to get the date formatted in YYYY-MM-DD format? Is it safe to rely on the CA locale using the ISO Date format? Locales can change over time, e.g. Canada switched to ISO Date format relatively recently. (It didn't use it [until at least 2016](https://www.cbc.ca/news/canada/proposed-legislation-aims-to-settle-date-debate-1.3407640).)

Based on [this StackOverflow answer](https://stackoverflow.com/a/78714460/69868) (that's not correct for formatting date-time), I think the following could work for us since we are interested only in the date part:

```js
const today = () => new Date(Date.now() - new Date().getTimezoneOffset() * 60000).toISOString().split('T')[0]
```

Thoughts?

stats/test/platform-routes.test.js
```diff
@@ -54,19 +58,7 @@ describe('Platform Routes HTTP request handler', () => {
...
-        { ...STATION_STATS, stationId: 'station2', acceptedMeasurementCount: 1, totalMeasurementCount: 1 }
-      ])
-      await givenDailyStationMetrics(pgPools.evaluate, '2024-01-12', [
-        { ...STATION_STATS, stationId: 'station2', acceptedMeasurementCount: 2, totalMeasurementCount: 2 },
-        { ...STATION_STATS, stationId: 'station3', acceptedMeasurementCount: 1, totalMeasurementCount: 1 }
-      ])
-      await givenDailyStationMetrics(pgPools.evaluate, '2024-01-13', [
-        { ...STATION_STATS, acceptedMeasurementCount: 1, totalMeasurementCount: 1 }
-      ])
+      await givenDailyMeasurementsSummary(pgPools.evaluate)
```
bajtos, line 61: I find this test difficult to reason about because it's not clear what data is created by `givenDailyMeasurementsSummary` and how it relates to what we assert.

At the same time, we need to keep the test setup brief, so that it's clear what data is important for this test and what's just "noise" needed by our DB schema.

I propose modifying the helper to allow the caller to provide only the column values they are interested in (`day` & `station_count` in this test). The helper should backfill any missing columns with reasonable defaults.

```js
await givenDailyMeasurementsSummary(pgPools.evaluate, [
  { day: '2024-01-10', station_count: 0 },
  { day: '2024-01-11', station_count: 1 },
  { day: '2024-01-12', station_count: 2 },
  { day: '2024-01-13', station_count: 3 }
])
```


Alternatively, we can keep the concept of `STATION_STATS`, which we were using before.

I prefer the helper to backfill any missing data with reasonable defaults.

stats/test/platform-routes.test.js
```diff
@@ -79,40 +71,25 @@ describe('Platform Routes HTTP request handler', () => {
       await assertResponseStatus(res, 200)
       const metrics = await res.json()
       assert.deepStrictEqual(metrics, [
-        { day: '2024-01-11', station_id_count: 1 },
-        { day: '2024-01-12', station_id_count: 2 }
+        { day: '2024-01-11', station_count: 3 },
+        { day: '2024-01-12', station_count: 4 }
```
bajtos, line 75: I believe the intention of this PR is to change the internal implementation while preserving the behaviour observed from outside. In that case it's best to keep the assertions unchanged, to show that the new implementation returns the same responses as the old one.

On the second look, I see that we are changing the response and renaming `station_id_count` to `station_count`. I think that's a good change, I am happy to land it. But I still think it's better to keep the count values unchanged.

The same comment applies to other tests below that may be changing the returned data.

stats/test/platform-routes.test.js
```diff
@@ -79,40 +71,25 @@ describe('Platform Routes HTTP request handler', () => {
...
-      await givenDailyStationMetrics(pgPools.evaluate, '2024-02-13', [
-        { ...STATION_STATS, acceptedMeasurementCount: 1, totalMeasurementCount: 1 }
-      ])
+      await givenMonthlyActiveStationCount(pgPools.evaluate, '2024-01-01', 3)
+      await givenMonthlyActiveStationCount(pgPools.evaluate, '2024-02-01', 1)
       // after the date range
-      await givenDailyStationMetrics(pgPools.evaluate, '2024-03-01', [
-        { ...STATION_STATS, acceptedMeasurementCount: 1, totalMeasurementCount: 1 }
-      ])
+      await givenMonthlyActiveStationCount(pgPools.evaluate, '2024-03-01', 1)
```
bajtos, line 88: Let's use a value that's different from February.

```suggestion
      await givenMonthlyActiveStationCount(pgPools.evaluate, '2024-03-01', 4)
```

stats/test/platform-routes.test.js
```diff
@@ -79,40 +71,25 @@ describe('Platform Routes HTTP request handler', () => {
...
 
       const res = await fetch(
         new URL(
-          '/stations/monthly?from=2024-01-11&to=2024-02-11',
+          '/stations/monthly?from=2024-01-01&to=2024-02-11',
```
bajtos, line 92: ❗ I think this is going to break the dashboard, and we need to preserve the original behavior.

In Grafana, we don't have a way to extract the month from the date; we always send the full date string. The REST API drops the day from the date string.

Take a look how `getMonthlyDistinctCount` does it using `date_trunc('month', value)`

```sql
    SELECT
      date_trunc('month', day)::DATE::TEXT as month,
      COUNT(DISTINCT ${safeColumn})::INT as ${safeAsColumn}
    FROM ${safeTable}
    WHERE
      day >= date_trunc('month', $1::DATE)
      AND day < date_trunc('month', $2::DATE) + INTERVAL '1 month'
    GROUP BY month
    ORDER BY month
```

stats/test/platform-routes.test.js
```diff
@@ -154,24 +119,45 @@ describe('Platform Routes HTTP request handler', () => {
       await assertResponseStatus(res, 200)
       const metrics = await res.json()
       assert.deepStrictEqual(metrics, [
-        { day: '2024-01-11', accepted_measurement_count: '1', total_measurement_count: '2' },
-        { day: '2024-01-12', accepted_measurement_count: '3', total_measurement_count: '4' }
+        { day: '2024-01-11', accepted_measurement_count: 11, total_measurement_count: 13 },
+        { day: '2024-01-12', accepted_measurement_count: 20, total_measurement_count: 22 }
```
bajtos, line 123: Same comments as above:

- The relevant test data should be explicitly expressed in the test (not hidden in helpers)
- Please setup the test data so that the request returns the same response data

bajtos (changes requested):
Great start!

Two most important topics:
- How to get the current date string that supports non-UTC locale
- Preserve support for `/stations/monthly?from=2024-01-11&to=2024-02-11`

PatrickNercessian:
> Can you update spark-evaluate version as part of this PR (similarly to what I did in #207)?

Sure. Do we also want to update the other mentions in package-lock.json? Or just the one from #207 ? There are also https://github.com/filecoin-station/spark-stats/blob/28fbe3a9b10c907f11052651d0f939f286397eec/package-lock.json#L5522 and https://github.com/filecoin-station/spark-stats/blob/28fbe3a9b10c907f11052651d0f939f286397eec/package-lock.json#L5560 which are pointing to an old commit.

stats/lib/platform-stats-fetchers.js
PatrickNercessian: I think I agree the name is not great, I think I just got used to it after the long planning stage and never reconsidered, my bad. Makes sense to change it before this PR lands.

We could also consider `daily_platform_stats_summary` or `daily_platform_summary`. It is a bit longer, but we refer to these as platform stats in spark-evaluate and spark-stats.

stats/test/platform-routes.test.js
PatrickNercessian: Agree with what you've said here, I'll switch the values back.

It was `station_id_count` before because we were using the dailyDistinctCount method which just appended the column name with "_count", but now I think `station_count` may be better.

The only real consumers of the data today, as far as I'm aware, is Grafana, and we can change the dashboard quite easily to match small naming changes like this. But if you'd prefer this PR to maintain exact external behavior, I can make those changes instead!

stats/test/platform-routes.test.js
PatrickNercessian: I'm sorry, I'm not sure I understand what would break the dashboard here?

The change to the test here was just so that we can test that it can grab both months correctly. If we prefer to do 2023-12-11 to 2024-02-11, I can do that. Not sure if that's what you mean though.

observer/test/observer.test.js
PatrickNercessian: This works for me! Will try it out

PatrickNercessian: Ah didn't notice your first comment in this thread. I was using it get the date formatted in YYYY-MM-DD, as suggested in that StackOverflow answer and elsewhere, but I do see the issue with potential future locale date format changes.

I personally prefer your second comment's idea as opposed to the DB query, but it's up to you!

stats/lib/platform-stats-fetchers.js
bajtos: I like `daily_platform_stats` most 👍🏻 

@juliangruber do you have any opinions?

stats/test/platform-routes.test.js
bajtos: +1 to what you wrote. Let's use `station_count` please.

stats/test/platform-routes.test.js
bajtos: I'd like to preserve the behaviour where `/stations/monthly?from=2024-01-11&to=2024-02-11` return data for both January and February.

Can you please revert the change on this line to verify that the new version handles this (edge?) case in the same way as the old version, and then add a new test for `/stations/monthly?from=2024-01-01&to=2024-02-11` if desired.

bajtos:
> > Can you update spark-evaluate version as part of this PR (similarly to what I did in #207)?
> 
> Sure. Do we also want to update the other mentions in package-lock.json? Or just the one from #207 ? There are also
> 
> https://github.com/filecoin-station/spark-stats/blob/28fbe3a9b10c907f11052651d0f939f286397eec/package-lock.json#L5522
> 
> and
> https://github.com/filecoin-station/spark-stats/blob/28fbe3a9b10c907f11052651d0f939f286397eec/package-lock.json#L5560
> 
> which are pointing to an old commit.

Thanks. I don't know which references to update, to be honest. In my experience, the combination of a git dependency and a package-lock file is not supported well by npm 😢 Perhaps you can update both, just to be sure.

stats/test/platform-routes.test.js
PatrickNercessian: Ah I see... Did not remember that `from=01-11` would also end up showing january data. I made changes to the test and to the query in https://github.com/filecoin-station/spark-stats/pull/210/commits/5362184846bbca36147ce4042a10ef6dcca1be76

stats/lib/platform-stats-fetchers.js
juliangruber: Since the outer functions only call the helper and nothing else, I would prefer to inline them, to reduce tech debt (the downside of the current solution is it takes you an unnecessary extra step to get to the business logic)

stats/lib/platform-stats-fetchers.js
juliangruber: +1 to changing the name before this PR lands, due to the coupling of the two repositories.

`daily_summary` is too generic - it's a summary of what? Unless it's a summary of the whole system. Is it that? Actually it could be. I'm +1 with this name if it is complete.

`daily_platform_stats` works for me 👍

observer/test/observer.test.js
bajtos: On the second though, I am fine with either your current implementation or the `getTimezoneOffset()` solution. I'd rather avoid making a DB roundtrip if we have a better option.

What's important is to add a code comment clearly spelling out:
1. What we are trying to achieve - get the current date in the current timezone in YYYY-MM-DD format
2. How the implementation works - if using your solution, explain that Canada uses the ISO date format.

stats/lib/platform-stats-fetchers.js
bajtos: in that case, let's do what @juliangruber wrote

stats/lib/platform-stats-fetchers.js
bajtos: Let's go with `daily_platform_stats` then.

juliangruber:
Once this lands, please follow up with https://github.com/filecoin-station/spark-stats/pull/215/files#diff-4236f3e7e6666ef758a371bb627b6810c7356fc347e81c1c197c5cb7b385bace

bajtos:
> Once this lands, please follow up with https://github.com/filecoin-station/spark-stats/pull/215/files#diff-4236f3e7e6666ef758a371bb627b6810c7356fc347e81c1c197c5cb7b385bace

It would be best to re-enable those tests as part of this pull request. I think the necessary changes in the production code are already part of this PR.

juliangruber (approved):

stats/lib/request-helpers.js
```diff
@@ -1,14 +1,14 @@
...
 
 /** @typedef {import('@filecoin-station/spark-stats-db').Queryable} Queryable */
 
 export const getDayAsISOString = (d) => d.toISOString().split('T')[0]
+export const todayDate = () => new Date(Date.now() - new Date().getTimezoneOffset() * 60000)
 
-export const today = () => getDayAsISOString(new Date())
-export const yesterday = () => getDayAsISOString(new Date(Date.now() - 24 * 60 * 60 * 1000))
+export const today = () => getDayAsISOString(todayDate())
+export const yesterday = () => getDayAsISOString(new Date(todayDate().getTime() - 24 * 60 * 60 * 1000))
```
bajtos, line 11: I find this confusing. We need to apply `getTimezoneOffset` only when converting the date-time value into `YYYY-MM-DD` format, could you please move this calculation to `getDayAsISOString`? That way the other places don't need any change.

```js
export const getDayAsISOString = (d) => new Date(d - new Date().getTimezoneOffset() * 60000).toISOString().split('T')[0]

export const today = () => getDayAsISOString(new Date())
export const yesterday = () => getDayAsISOString(new Date(Date.now() - 24 * 60 * 60 * 1000))
```

Thoughts?

stats/lib/request-helpers.js
```diff
@@ -89,7 +89,7 @@ export const getStatsWithFilterAndCaching = async (pathname, pathParams, searchP
 const setCacheControlForStatsResponse = (res, filter) => {
   // We cannot simply compare filter.to vs today() because there may be a delay in finalizing
   // stats for the previous day. Let's allow up to one hour for the finalization.
-  const boundary = getDayAsISOString(new Date(Date.now() - 3600_000))
+  const boundary = getDayAsISOString(new Date(todayDate().getTime() - 3600_000))
```
bajtos, line 92: Oh, I see now why you need `todayDate`.

Let's find more descriptive names for `todayDate` and `getDayAsISOString` then. The names should convey that:

- `todayDate` is a date-time value which, when converted to UTC timezone, produces the same date (day) as is in the local timezone now.
- `getDayAsISOString` returns the date part of the provided parameter expressed in the UTC timezone

That sounds too complex to me. Maybe we are approaching this from a wrong angle.

What we want to implement:
- If we are inside the interval `0:00:00-1:00:00` in the local timezone, then we compare `filter.to` against _yesterday_ date.
- Otherwise, we compare `filter.to` against _today_ date.

How about converting `filter.to` to a full date-time value (in local timezone) and the comparing it against `new Date()`?

```js
// Convert "2024-09-01" to "2024-09-02T01:00:00.000" in local timezone
const adjustedFilterTo = new Date(
  new Date(filter.to) + 25 * 3600_000 
)
if (adjustedFilterTo >= new Date()) {
  // response includes partial data for today, cache it for 10 minutes only
// etc.
```

bajtos (changes requested):
Great progress, @PatrickNercessian! The PR is almost good to be landed.

stats/lib/request-helpers.js
juliangruber: I had assumed this part had been thoroughly discussed by you two, so I accepted my confusion and thought there would be good reasons.

+1 to what @bajtos said

stats/lib/request-helpers.js
PatrickNercessian: > todayDate is a date-time value which, when converted to UTC timezone, produces the same date (day) as is in the local timezone now.

I think this is not true, unless I am misunderstanding. `todayDate` is simply the local date based on the timezone of the computer running the code, as far I understand it. Maybe we can change the name to `todayLocalDate`?

> getDayAsISOString returns the date part of the provided parameter expressed in the UTC timezone

I think this name is fine? Its functionality is unchanged from before in the current form of this PR, and describes its behavior accurately I think. It does not have any behavior related to UTC timezone, not sure if I am misunderstanding your statement here.

> How about converting filter.to to a full date-time value (in local timezone) and the comparing it against new Date()?

I don't know, to me this seems harder to understand than the PR as it is now. But I'll default to whatever you and/or Julian decide on.

bajtos: I think the current solution have so many subtle nuances and complexities that we should step back and look for better options.

What we are trying to achieve: Given a `Date` value in the local time zone, we want to get a date string in the YYYY-MM-DD format for the local day.

How about this solution:

- Create a function called `getLocalDayAsISOString`, to make it clear this function is preserving the local timezone
- Implement the function by calling year/month/day getters, padding the values to 2/4 digits and concatenating them to the desired format.

```js
const getLocalDayAsISOString = (d) => {                                         
  return [                                                                      
    d.getFullYear(),                                                            
    String(d.getMonth() + 1).padStart(2, '0'),                                  
    String(d.getDate()).padStart(2, '0')                                        
  ].join('-')                                                                   
}
```

This function is very explicit about what it does. I think it should also not require any changes in how we calculate the values passed in the `d` parameter.

At the top of this file:

```js
export const today = () => getLocalDayAsISOString(new Date())
export const yesterday = () => getLocalDayAsISOString(new Date(Date.now() - 24 * 60 * 60 * 1000))
```

In the code we are discussing in this thread:

```js
const boundary = getLocalDayAsISOString(new Date(Date.now() - 3600_000))
```

@juliangruber @PatrickNercessian WDYT?

db/index.js
```diff
@@ -109,3 +110,20 @@ export const migrateStatsDB = async (client) => {
 
   console.log('Migrated `spark-stats` DB schema to version', await postgrator.getDatabaseVersion())
 }
+
+/**
+ * @param {import('./typings.js').Queryable} pgPool
+ * @param {string} day
+ * @param {string[]} participantAddresses
+ */
+export const givenDailyParticipants = async (pgPool, day, participantAddresses) => {
```
bajtos, line 119: `db/index.js` is a production code. `getLocalDayAsISOString ` is a test-data-builder helper used in tests; it does not belong to the production code.

I see that `getLocalDayAsISOString` is shared by both `observer/test` and `stats/test`; maybe that's the reason why we decided it's best to move it to spark-evaluate.

Anyhow, let's not overengineer this.

I propose creating a new file, `db/test-helpers.js`, and putting the helper function there.

stats/lib/request-helpers.js
juliangruber: +1 to being very explicit here. The previous version to me also is too magic.

Could the proposed version be simplified to this?

```js
const getLocalDayAsISOString = (d) => {
  return d.toISOString().split('T')[0]
}
```

I believe we had this version before, so I think this is _not_ what we need, which also tells me that the version @bajtos proposed isn't going to work. @PatrickNercessian what are we missing?

db/index.js
PatrickNercessian: That was my first (actually second, I first attempted to put it in `stats` instead of `db`) attempt, but I was unable to get the other packages to recognize the file if it was not in index.js. See https://github.com/filecoin-station/spark-stats/pull/210/commits/865e0f158ffb75918b60b08bb72fc92a859dd5a8 and https://github.com/filecoin-station/spark-stats/pull/210/commits/81addc2f3559b243144c5c3e222d0047e40afcdb. Any ideas on what needs to be done to make it work?

PatrickNercessian: Nice, it worked! When I had originally tried it, I must have done `git add -u .` as I usually do, and forgot that I had a new file that needed to be added. Thoughtless mistake, my bad!

stats/lib/request-helpers.js
juliangruber: > @juliangruber I think the issue with that one is that it is does not preserve the local date, which was the original problem.

Right, `d.toISOString()` is in UTC.

@bajtos's version works, as Date methods like `getDay()` return according to local time (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay).

I'm in favor of using that. And I'm happy I learned something new about JS Date functions

bajtos: I feel we are spending too much time on a very minor issue. What's the simplest way to resolve this?

Here are two options I like:

1. We can rule that **the DB must run on a computer configured with the UTC timezone**. This matches our production system, GitHub Actions, and some Docker deployments of Postgres. In this case, I believe the original implementation `d.toISOString().split('T')[0]` works best. However, this condition is not met when the DB runs as a local service on the dev computer - @PatrickNercessian's case.

2. Alternatively, we can rule that **the DB must run in the same timezone as the Node.js processes**. This matches our production system, GitHub Actions, and a database server running as a local service. In this case, I believe the string-manipulation based solution I proposed in https://github.com/filecoin-station/spark-stats/pull/210#discussion_r1756908718 works best. However, this condition is not met when the DB runs via Docker on the dev computer - @bajtos's case.

What happens when the condition is not met: there is a window around midnight UTC when the tests fail when executed locally. 

IMO, the second option - "DB and Node.js processes have the same timezone" - is what most people would assume, and it's also the easiest to reason about. Therefore, I prefer to go with that.

Personally, I will either figure out how to run the DB in my local timezone, or, and this is more likely, I'll remember that the tests will fail when I run them between midnight and 2am CEST, and ignore the problem for now.

> All of this depends on the postgres configuration, so I think we must do an SQL query for this. If we switch javascript to use local date, and PostGres is set to UTC (as in your case), tests will also fail at certain times.

This adds an extra query to our REST API endpoints. IIUC, its only benefit is solving an edge case on developer machines only. I'd rather avoid paying that performance cost.

PatrickNercessian: Gotcha. Made the change to use the string-manipulation solution you suggested. However, the downside it's duplicated logic, in both `stats` and `observer`, since they both require the functionality.

We could always put them in `db`, such as what we did with the test-helper `givenDailyParticipants`, but this seems even less related to the database. And it's also not a test-helper really, since it's also used in the production code.

So I ended up duplicating them. Let me know if you'd prefer some other solution

juliangruber: To me this method isn't big enough to require deduplication

bajtos: I am fine with duplication, too.

bajtos:
I think the dry-run script is failing because it does not have access to GLIF_TOKEN. Nothing to worry about.

```
Querying scheduled rewards from impact evaluator
JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
file:///home/runner/work/spark-stats/spark-stats/node_modules/ethers/lib.esm/utils/errors.js:124
            error = new Error(message);
                    ^
Error: server response 401 Unauthorized (request={  }, response={  }, error=null, info={ "requestUrl": "https://api.node.glif.io/rpc/v0", "responseBody": "<html>\r\n<head><title>401 Authorization Required</title></head>\r\n<body>\r\n<center><h1>401 Authorization Required</h1></center>\r\n</body>\r\n</html>\r\n", "responseStatus": "401 Unauthorized" }, code=SERVER_ERROR, version=6.13.2)
```

PR #221, opened 2024-09-11
Get recent participants from `spark-evaluations-recent-participants`

juliangruber:
This removes one db coupling between spark-stats and spark-evaluate


bajtos (commented):
Is this still relevant?

PR #234, opened 2024-10-07
Add integrate `spark-rewards`

juliangruber:
https://github.com/space-meridian/roadmap/issues/161

### DIFF
--- a/observer/lib/observer.js
+++ b/observer/lib/observer.js
@@ -1,5 +1,6 @@
 import { updateDailyTransferStats } from './platform-stats.js'
 import * as Sentry from '@sentry/node'
+import assert from 'node:assert'
 
 /**
  * Observe the transfer events on the Filecoin blockchain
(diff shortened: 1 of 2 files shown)


bajtos (approved):
LGTM.

I don't see any tests for this change, what's your plan to verify the code works as intended?

observer/lib/observer.js
```diff
@@ -35,8 +35,24 @@ export const observeTransferEvents = async (pgPoolStats, ieContract, provider) =
...
 }
 
+const getScheduledRewards = async (address, ieContract) => {
+  const [fromContract, fromSparkRewards] = await Promise.all([
+    ieContract.rewardsScheduledFor(address),
+    (async () => {
+      const res = await fetch(
+        `https://spark-rewards.fly.dev/scheduled-rewards/${address}`
+      )
+      const json = await res.json()
```
bajtos, line 45: How do you handle 4xx and 5xx responses? What if the server responds with 5xx and the response body provides error details? Should you check `res.ok`?

In other places, I had a great experience with the helper `assertOkResponse` that reads the error response body and includes it in `err.message`.

juliangruber: see reply to same question in other PR

bajtos: Link to that reply: https://github.com/filecoin-station/core/pull/606#discussion_r1791853853

PR #244, opened 2024-10-31
Add daily retrieval result codes

juliangruber:


README.md
```diff
@@ -85,6 +84,10 @@ Base URL: http://stats.filspark.com/
 
   http://stats.filspark.com/deals/summary
 
+- `GET /retrieval-result-status/daily?from=2024-01-01&to=2024-01-31`
```
bajtos, line 87: I don't understand the "status" part.

In the docs (see https://www.notion.so/protocollabs/Spark-Request-Based-Non-Committee-Global-Retrieval-Success-Rate-4c5e8c47c45f467f80392d00cac2aae4?pvs=4#122837df73d4803f917bf8e8eb13f9d4), we use the term "Retrieval Result Code". 

Can we use that term for the REST API too?

```suggestion
- `GET /retrieval-result-codes/daily?from=2024-01-01&to=2024-01-31`
```

db/migrations/005.do.retrieval-result-status.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE daily_retrieval_result_status (
```
bajtos, line 1: Same here?

```suggestion
CREATE TABLE daily_retrieval_result_codes (
```

observer/lib/observer.js
```diff
@@ -96,3 +96,28 @@ export const observeScheduledRewards = async (pgPools, ieContract, fetch = globa
...
 }
+
+export const observeRetrievalResultStatus = async (pgPoolStats, influxQueryApi) => {
+  const rows = await influxQueryApi.collectRows(`
+    import "strings"
+    from(bucket: "spark-evaluate")
+      |> range(start: 0)
+      |> filter(fn: (r) => r["_measurement"] == "retrieval_stats_honest")
+      |> filter(fn: (r) => strings.hasPrefix(v: r._field, prefix: "result_rate_"))
+      |> aggregateWindow(every: 1d, fn: mean, createEmpty: false)
```
bajtos, line 107: I _think_ this will likely produce slightly incorrect data.

My understanding:

1. spark-evaluate creates a new set of `result_rate_*` datapoints every round. Each `result_rate_*` value is a fraction of how many measurements reported that result code.
2. In your query, you are aggregating these per-round data rates to per-day rates by calculating the mean average of the fractions.

Consider the following case (simplified):

1. On round 1, we collect 400k measurements, of which 40k report OK. result_rate_OK is 10%.
2. On round 2, we collect 600k measurements, of which 50k report OK. result_rate_OK is 8.33%.

Mean average: (10+8.33)/2 = 9.165%
Precise value: (40k+50k)/(400k+600k) = 9%

observer/lib/observer.js
```diff
@@ -96,3 +96,28 @@ export const observeScheduledRewards = async (pgPools, ieContract, fetch = globa
 function isEventLog (logOrEventLog) {
   return 'args' in logOrEventLog
 }
+
+export const observeRetrievalResultStatus = async (pgPoolStats, influxQueryApi) => {
+  const rows = await influxQueryApi.collectRows(`
+    import "strings"
+    from(bucket: "spark-evaluate")
+      |> range(start: 0)
```
bajtos, line 104: Keep in mind we have a 90-day retention for the spark-evaluate InfluxDB bucket. Is your solution going to handle this?

It also seems to be a bit inefficient to always query all 90 days when in most cases, we need only data for yesterday.

Having said that, I guess we don't worry about the performance (and the InfluxDB bill) right now.

observer/test/observer.test.js
```diff
@@ -184,4 +184,27 @@ describe('observer', () => {
...
+
+  describe('observeRetrievalResultStatus', () => {
+    beforeEach(async () => {
+      await pgPools.stats.query('DELETE FROM daily_retrieval_result_status')
+    })
+
+    it('observes retrieval result status', async () => {
+      await observeRetrievalResultStatus(pgPools.stats, {
+        collectRows: async () => [
+          { _time: today(), _field: 'success', _value: 0.5 },
```
bajtos, line 196: Can you please use real field names and values, so that the test better documents what we expect on the other side of this integration.

db/migrations/005.do.retrieval-result-status.sql
```diff
@@ -0,0 +1,6 @@
...
+  day DATE NOT NULL,
+  status TEXT NOT NULL,
+  rate NUMERIC NOT NULL,
+  PRIMARY KEY (day, status)
+);
```
bajtos, line 6: Since you are going to query this table using a `day` range, please add an index to speed up those queries.

stats/test/handler.test.js
```diff
@@ -434,6 +435,37 @@ describe('HTTP request handler', () => {
...
+    it('returns daily retrieval result status for the given date range', async () => {
+      await pgPools.stats.query(`
+        INSERT INTO daily_retrieval_result_status
+        (day, status, rate)
+        VALUES
+        ('2024-01-11', 'Error1', 0.1),
+        ('2024-01-11', 'Error2', 0.9),
+        ('2024-01-12', 'Error1', 1),
+        ('2024-01-13', 'Error1', 0.5),
+        ('2024-01-13', 'Error3', 0.5)
```
bajtos, line 448: Ditto, please use real retrieval result codes and rates.

observer/lib/observer.js
bajtos: What we are doing in other places:
- In spark-evaluate, we keep track of both per-result count and total count
- In spark-stats, we calculate the % rate from those numbers

For example: per-miner OK results:
- https://github.com/filecoin-station/spark-evaluate/blob/170e86e69fd038594df876332898138225b6bd35/lib/public-stats.js#L42-L57
- https://github.com/filecoin-station/spark-stats/blob/09db04a89749339f3b3ac7f53218054e149940f8/stats/lib/stats-fetchers.js#L6-L26

README.md
juliangruber: Good idea!

juliangruber: #18763e9

db/migrations/005.do.retrieval-result-status.sql
juliangruber: 18763e9

observer/lib/observer.js
juliangruber: > Keep in mind we have a 90-day retention for the spark-evaluate InfluxDB bucket. Is your solution going to handle this?

This solution works precisely because we have the 90 day retention. Without it, the query would be unbounded.

> It also seems to be a bit inefficient to always query all 90 days when in most cases, we need only data for yesterday.
> Having said that, I guess we don't worry about the performance (and the InfluxDB bill) right now.

I agree, I'm going for the pragmatic solution here.

observer/test/observer.test.js
juliangruber: f2a7587

db/migrations/005.do.retrieval-result-status.sql
juliangruber: caf8af7

stats/test/handler.test.js
juliangruber: f75b30b

observer/lib/observer.js
juliangruber: Good point, the mean over relative values will be off with different measurement count. Since measurement count is constant-ish now, I think this doesn't matter a lot. But it should still be improved.

Do you have a suggestion how to solve this with Influx queries? I assume we'd need to join over the measurement count.

I have added a comment in 818285d, are you ok shipping like this?

bajtos: > Do you have a suggestion how to solve this with Influx queries? I assume we'd need to join over the measurement count.

I don't think you can solve this at the query level. Maybe you could write a query that multiplies each `result_rate_XXX` value with `measurements` (total count) to reconstruct the number of measurements that reported the result code `XXX`.

IMO, `result_rate_XXX` points are not the right data source for this feature. We can keep pushing in this direction to see if we can find a fix. It may be better to step back and add a new stats writer to spark-evaluate, as we already do for other stats.

> I have added a comment in https://github.com/filecoin-station/spark-stats/commit/818285dc7f44d84559afa5d2aecefff38df03c00, are you ok shipping like this?

What are the ramifications of shipping like this:
- The chart will show incorrect data. (Hopefully, the inaccuracy is too small for anybody to notice. But we don't know how bad it will be, do we?)
- We are introducing a technical debt. In the worst-case scenario, we will need to remove the InfluxDB observer and implement something else. Since we are talking about ~200 LOC, I think this is not a big deal.

While I am not excited about shipping it like this, I think it's an acceptable trade-off to make. 👍🏻

bajtos (approved):

observer/lib/observer.js
bajtos: > > Keep in mind we have a 90-day retention for the spark-evaluate InfluxDB bucket. Is your solution going to handle this?
> 
> This solution works precisely because we have the 90 day retention. Without it, the query would be unbounded.

Would you mind adding a code comment to explain this to future readers?

juliangruber: c62756a

observer/lib/observer.js
juliangruber: > I don't think you can solve this at the query level. Maybe you could write a query that multiplies each result_rate_XXX value with measurements (total count) to reconstruct the number of measurements that reported the result code XXX.

Yes this is what I was thinking about. But this join is going to be expensive.

> IMO, result_rate_XXX points are not the right data source for this feature. We can keep pushing in this direction to see if we can find a fix. It may be better to step back and add a new stats writer to spark-evaluate, as we already do for other stats.

Yeah post LabWeek we can work on the proper solution for these.

> What are the ramifications of shipping like this:
> 
> * The chart will show incorrect data. (Hopefully, the inaccuracy is too small for anybody to notice. But we don't know how bad it will be, do we?)

I propose let's first see how this looks in the dashboard and then decide if it's too wrong to ship.

> * We are introducing a technical debt. In the worst-case scenario, we will need to remove the InfluxDB observer and implement something else. Since we are talking about ~200 LOC, I think this is not a big deal.

The worst case is one revert away so I think this is ok. Also, from a business perspective, if we're asked about why we didn't add this chart that was requested last LabWeek, technical debt is not a good answer. Do you have a realistic alternative proposal?

> While I am not excited about shipping it like this, I think it's an acceptable trade-off to make. 👍🏻

I'm at the moment not operating at the level of excitement :D I just want to ship (obviously without shipping something too bad) so we have more useful things to show.

PR #248, opened 2024-11-13
Add GET /participants/summary

juliangruber:
For https://github.com/space-meridian/roadmap/issues/154.

<del>@bajtos currently tests are failing, since any request without from/to query parameters gets redirected to one with. This filter is meaningless in the context of this route, which currently exposes the total seen participant count. Do you have a proposal what to do about this?</del>

### DIFF
--- a/.github/workflows/dependabot-auto-approve-minor.yml
+++ b/.github/workflows/dependabot-auto-approve-minor.yml
@@ -20,6 +20,7 @@ jobs:
           - postgrator
           - '@types/'
           - standard
+          - cross-spawn
     steps:
       - name: Dependabot metadata
         id: metadata
--- a/observer/package.json
+++ b/observer/package.json
@@ -18,7 +18,7 @@
     "@filecoin-station/spark-impact-evaluator": "^1.1.1",
     "@filecoin-station/spark-stats-db": "^1.0.0",
     "@influxdata/influxdb-client": "^1.35.0",
-    "@sentry/node": "^8.38.0",
+    "@sentry/node": "^8.39.0",
     "debug": "^4.3.7",
     "ethers": "^6.13.4",
     "pg": "^8.13.1"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1524,21 +1524,21 @@
       "license": "BSD-3-Clause"
     },
     "node_modules/@sentry/core": {
-      "version": "8.38.0",
-      "resolved": "https://registry.npmjs.org/@sentry/core/-/core-8.38.0.tgz",
-      "integrity": "sha512-sGD+5TEHU9G7X7zpyaoJxpOtwjTjvOd1f/MKBrWW2vf9UbYK+GUJrOzLhMoSWp/pHSYgvObkJkDb/HwieQjvhQ==",
+      "version": "8.39.0",
+      "resolved": "https://registry.npmjs.org/@sentry/core/-/core-8.39.0.tgz",
+      "integrity": "sha512-rg2mHtwdCaedqub7bd+ht08vZgtwPO7el5m5sPNeb7V75GcQwSziu6G02vGxCBCsAHpoFn1A+0JLEajaYzZI7w==",
       "dependencies": {
-        "@sentry/types": "8.38.0",
-        "@sentry/utils": "8.38.0"
+        "@sentry/types": "8.39.0",
+        "@sentry/utils": "8.39.0"
       },
       "engines": {
         "node": ">=14.18"
       }
     },
     "node_modules/@sentry/node": {
-      "version": "8.38.0",
-      "resolved": "https://registry.npmjs.org/@sentry/node/-/node-8.38.0.tgz",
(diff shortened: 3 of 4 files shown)


bajtos:
> @bajtos currently tests are failing, since any request without from/to query parameters gets redirected to one with. This filter is meaningless in the context of this route, which currently exposes the total seen participant count. Do you have a proposal what to do about this?

Sure! Don't use the `respond` helper that's adding `from/to` redirects & caching, implement the new endpoint as a raw HTTP route.

See e.g. here:
https://github.com/filecoin-station/spark-stats/blob/5404dd1e4a300c1251005a987dea7180ef98bb3f/stats/lib/handler.js#L108-L121

A mock-up to illustrate what I mean:

```js
// router:

 } else if (req.method === 'GET' && url === '/participants/summary') {
    await getParticipantsSummary(req, res, pgPools.stats)
 } // etc.


// implementation

// todo: add types
export const gatParticipantsSummary = async (req, res, pgPool) => {
  const { rows } = await pgPool.query(`
    SELECT COUNT(DISTINCT participant_id) FROM daily_participants
  `)

  // TODO: configure caching!

  json(res, {
    participant_count: rows[0].count
  })
}
```

stats/lib/platform-routes.js
```diff
@@ -36,6 +38,8 @@ export const handlePlatformRoutes = async (req, res, pgPools) => {
     await respond(pgPools.evaluate, fetchParticipantsWithTopMeasurements)
   } else if (req.method === 'GET' && url === '/participants/top-earning') {
     await respond(pgPools.stats, fetchTopEarningParticipants)
+  } else if (req.method === 'GET' && url === '/participants/summary') {
+    json(res, await fetchParticipantsSummary(pgPools.evaluate))
```
bajtos, line 42: Hang on. The SQL query takes more than 400ms. I am not comfortable running that query on every request; we need to cache the result.

bajtos (changes requested):

stats/lib/platform-routes.js
juliangruber: 410b9cc

stats/lib/platform-routes.js
```diff
@@ -39,7 +39,9 @@ export const handlePlatformRoutes = async (req, res, pgPools) => {
   } else if (req.method === 'GET' && url === '/participants/top-earning') {
     await respond(pgPools.stats, fetchTopEarningParticipants)
   } else if (req.method === 'GET' && url === '/participants/summary') {
-    json(res, await fetchParticipantsSummary(pgPools.evaluate))
+    const summary = await fetchParticipantsSummary(pgPools.evaluate)
+    res.setHeader('cache-control', `public, max-age=${24 * 3600 /* one day */}`)
+    json(res, summary)
```
bajtos, line 44: Would you mind extracting these three lines into a new function? I would like to keep `handlePlatformRoutes` at a single level of abstraction - match request method+path and call the appropriate route handler.

juliangruber: I have refactored dc18234 but don't know where to put it. I also don't see a good way to refactor a generic `getStatsWithCaching` function. Can you please advise in more detail how you envision this?

juliangruber:
@bajtos please see my comment above: I don't know a good place to put the new function. Since it's not a data fetcher but a responder, it shouldn't be in the fetchers file

bajtos:
> @bajtos please see my comment above: I don't know a good place to put the new function. Since it's not a data fetcher but a responder, it shouldn't be in the fetchers file

I see what you mean now.

Since the new function is a route handler, I would put it at the bottom of `stats/lib/platform-routes.js`, below the main routing function. 

We are already using that pattern in [stats/lib/handler.js](https://github.com/filecoin-station/spark-stats/blob/main/stats/lib/handler.js).

bajtos:
Yeah, I find this new version harder to read too.

I was thinking about what I described in https://github.com/filecoin-station/spark-stats/pull/248#issuecomment-2482535064

> ```js
> // router:
> 
>  } else if (req.method === 'GET' && url === '/participants/summary') {
>     await respondWithParticipantsSummary(req, res, pgPools.stats)
>  } // etc.

To me, I like the router to be just a router delegating to other functions, with as little configuration as necessary.

Then, the route handler function (`respondWithParticipantsSummary`) goes one level down the abstraction hierarchy and deals with details like TTL duration for the cache.

TBH, I think we are hitting diminishing returns here. I am fine with [dc18234](https://github.com/filecoin-station/spark-stats/pull/248/commits/dc182345221fd98c67de3eb83e3652245a5138dc). I consider [b1dd0f2](https://github.com/filecoin-station/spark-stats/pull/248/commits/b1dd0f26404950ebcb82864e8603670935823694) a bit worse, but I can live with that too.

This code is very isolated and should be super easy to change in the future if needed.

Unless you are concerned about setting up the right patterns for future contributors, in which case, I prefer the option I described at the top of this comment.

bajtos (commented):
👏🏻

PR #256, opened 2024-11-21
feat: telemetry - loop duration & delay

bajtos:
Monitor how long each loop takes to finish and what's the delay before the next
iteration, so that we can alert when things become too slow.

### DIFF
--- a/observer/bin/spark-observer.js
+++ b/observer/bin/spark-observer.js
@@ -3,7 +3,7 @@ import * as SparkImpactEvaluator from '@filecoin-station/spark-impact-evaluator'
 import { ethers } from 'ethers'
 import * as Sentry from '@sentry/node'
 import timers from 'node:timers/promises'
-import { InfluxDB } from '@influxdata/influxdb-client'
+import { createInflux } from '../lib/telemetry.js'
 import assert from 'node:assert/strict'
 
 import { RPC_URL, rpcHeaders } from '../lib/config.js'
@@ -26,11 +26,7 @@ const provider = new ethers.JsonRpcProvider(fetchRequest, null, { polling: true 
 
 const ieContract = new ethers.Contract(SparkImpactEvaluator.ADDRESS, SparkImpactEvaluator.ABI, provider)
 
-const influx = new InfluxDB({
-  url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
-  // spark-stats-observer-read
-  token: INFLUXDB_TOKEN
-})
+const { influx, recordTelemetry } = createInflux(INFLUXDB_TOKEN)
 
 const influxQueryApi = influx.getQueryApi('Filecoin Station')
 
@@ -47,7 +43,14 @@ const loop = async (name, fn, interval) => {
     }
     const dt = Date.now() - start
     console.log(`Loop "${name}" took ${dt}ms`)
-    await timers.setTimeout(interval - dt)
+    recordTelemetry(`loop_${name.replaceAll(' ', '_')}`, point => {
+      point.intField('interval_ms', interval)
+      point.intField('duration_ms', dt)
+      point.intField('delay_ms', interval - dt)
+    })
+    if (dt < interval) {
+      await timers.setTimeout(interval - dt)
+    }
   }
 }
 
--- a/observer/bin/spark-observer.js
+++ b/observer/bin/spark-observer.js
@@ -5,6 +5,7 @@ import * as Sentry from '@sentry/node'
 import timers from 'node:timers/promises'
 import { createInflux } from '../lib/telemetry.js'
 import assert from 'node:assert/strict'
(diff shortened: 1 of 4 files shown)


observer/bin/spark-observer.js
```diff
@@ -47,7 +43,14 @@ const loop = async (name, fn, interval) => {
     }
     const dt = Date.now() - start
     console.log(`Loop "${name}" took ${dt}ms`)
-    await timers.setTimeout(interval - dt)
+    recordTelemetry(`loop_${name.replaceAll(' ', '_')}`, point => {
+      point.intField('interval_ms', interval)
+      point.intField('duration_ms', dt)
+      point.intField('delay_ms', interval - dt)
+    })
+    if (dt < interval) {
```
juliangruber, line 51: This condition isn't necessary as `setTimeout(-10)` has the same result as `setTimeout(0)`. Up to you

observer/bin/spark-observer.js
```diff
@@ -47,7 +43,14 @@ const loop = async (name, fn, interval) => {
     }
     const dt = Date.now() - start
     console.log(`Loop "${name}" took ${dt}ms`)
-    await timers.setTimeout(interval - dt)
+    recordTelemetry(`loop_${name.replaceAll(' ', '_')}`, point => {
+      point.intField('interval_ms', interval)
+      point.intField('duration_ms', dt)
+      point.intField('delay_ms', interval - dt)
+    })
```
bajtos, line 50: I am envisioning an alert like "if delay_ms/interval_ms is less than 10%"

observer/bin/spark-observer.js
```diff
@@ -47,7 +43,14 @@ const loop = async (name, fn, interval) => {
     }
     const dt = Date.now() - start
     console.log(`Loop "${name}" took ${dt}ms`)
-    await timers.setTimeout(interval - dt)
+    recordTelemetry(`loop_${name.replaceAll(' ', '_')}`, point => {
```
juliangruber, line 46: `name` is meant to be a human readable string, not an identifier. I think we should either add a telemetry identifier `{ id: 'transfer-events' }` or use a safe method of converting that doesn't produce failures if we use a name like `"oh:what?!"`, like https://www.npmjs.com/package/slugify

juliangruber (changes requested):

observer/bin/spark-observer.js
bajtos: Well, maybe "if delay_ms/interval_ms is less than 40%" is better as it will give us more time to fix the issue before it becomes a problem.

observer/bin/spark-observer.js
bajtos: I was not aware of that. I replaced the condition with a code comment explaining this Node.js feature.

juliangruber: It's not a Node.js feature apparently, it's part of the spec: https://stackoverflow.com/a/8431015

I find the comment worse than the condition (it makes me think more). Would you be ok with reverting to the condition?

observer/bin/spark-observer.js
```diff
@@ -47,6 +44,14 @@ const loop = async (name, fn, interval) => {
     }
     const dt = Date.now() - start
     console.log(`Loop "${name}" took ${dt}ms`)
+
+    recordTelemetry(`loop_${slug(name, '_')}`, point => {
+      point.intField('interval_ms', interval)
+      point.intField('duration_ms', dt)
+      point.intField('delay_ms', interval - dt)
```
juliangruber, line 51: This field is redundant as it can easily be computed through the other ones
```suggestion
```

juliangruber (changes requested):

observer/bin/spark-observer.js
bajtos: Sure, no problem, I'll use the condition.

observer/bin/spark-observer.js
bajtos: In the past, I had bad experience with using computed fields in Grafana, but it was probably a different situation. I am fine to remove this field now, and bring it back only if I need it.

juliangruber (approved):

bajtos:
I created a new Grafana panel, configured an alert when the iteration takes longer than 60% of the interval between iterations, and disabled the CPU throttling alert.

Screenshot 2024-11-27 at 15 11 08

bajtos:
Unfortunately, the alert does not work 😠 

The composer UI shows that all is good. But when the Grafana alerting system runs, it reports the following error:

```
Something went wrong when evaluating this alert rule
invalid format of evaluation results for the alert definition : frame cannot uniquely be identified by its labels: has duplicate results with labels {}
```

I searched the internet but could not find any useful answer. This is such a waste of time. I decided to re-enable the cpu-throttling alert and pause the new one.

PR #257, opened 2024-11-21
feat: enable Sentry Performance Profiling

bajtos:
Setup Sentry Performance Profiling for spark-stats to allow us to learn why
are we hitting Fly's CPU throttling limits. Depending on the outcome of this
experiment, we can roll out profiling to other services, e.g. spark-observer.

Sentry docs:
- https://docs.sentry.io/platforms/javascript/guides/node/tracing/
- https://docs.sentry.io/platforms/javascript/guides/node/tracing/instrumentation/automatic-instrumentation/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1647,6 +1647,26 @@
         "node": ">=10"
       }
     },
+    "node_modules/@sentry/profiling-node": {
+      "version": "8.39.0",
+      "resolved": "https://registry.npmjs.org/@sentry/profiling-node/-/profiling-node-8.39.0.tgz",
+      "integrity": "sha512-0/xHWEkB40IRUs5p3UAdIw28HxG9j1vprWno9SLKXZ2e7zNpz7MZmtOKdiXcs/uzOCcxW069sENf9uue5WqOxA==",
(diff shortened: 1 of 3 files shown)


juliangruber (approved):

juliangruber:
Let's watch CPU usage after deploying

bajtos:
The CPU usage seems to be fine.

However, it seems that spark-stats is not reporting any performance information to Sentry 😢

bajtos:
https://docs.sentry.io/platforms/javascript/guides/node/profiling/#enable-continuous-profiling

> The current profiling implementation stops the profiler automatically after 30 seconds (unless you manually stop it earlier). Naturally, this limitation makes it difficult to get full coverage of your app's execution. We now offer an experimental continuous mode, where profiling data is periodically uploaded while running, with no limit on how long the profiler may run.
>
> These new APIs do not offer any sampling functionality—every call to start the profiler will run and start sending profiling data. If you are interested in reducing the amount of profiles that run, you must take care to do it at the callsites.
>
> Continuous profiling has implications for your org's billing structure.

I think we need to research the above ☝🏻

PR #267, opened 2024-12-10
Add accumulative daily participants

juliangruber:
For https://github.com/space-meridian/roadmap/issues/154#issuecomment-2511707545

https://chat.mistral.ai/chat/4c43d4a1-2126-4e17-b65c-8ba40afac14a

I modified to add `DISTINCT(day::TEXT)`, as we were otherwise getting redundant rows. I also changed `cumulative_participants` to an `INT` and renamed it to `participants`

### DIFF
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -83,17 +83,10 @@ export const fetchDailyRewardTransfers = async (pgPool, filter) => {
  */
 export const fetchAccumulativeDailyParticipantCount = async (pgPool, filter) => {
   const { rows } = await pgPool.query(`
-    WITH daily_participants_count AS (
-      SELECT
-        day,
-        participant_id,
-        ROW_NUMBER() OVER (PARTITION BY participant_id ORDER BY day) AS rn
+    WITH first_appearance AS (
+      SELECT participant_id, MIN(day) as day
       FROM daily_participants
-    ),
-    first_appearance AS (
-      SELECT day, participant_id
-      FROM daily_participants_count
-      WHERE rn = 1
+      GROUP BY participant_id
     ),
     cumulative_participants AS (
       SELECT
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -83,10 +83,29 @@ export const fetchDailyRewardTransfers = async (pgPool, filter) => {
  */
 export const fetchAccumulativeDailyParticipantCount = async (pgPool, filter) => {
   const { rows } = await pgPool.query(`
-    SELECT day::TEXT, COUNT(DISTINCT participant_id) as participants
-    FROM daily_participants
+    WITH daily_participants_count AS (
+      SELECT
+        day,
+        participant_id,
+        ROW_NUMBER() OVER (PARTITION BY participant_id ORDER BY day) AS rn
+      FROM daily_participants
+    ),
+    first_appearance AS (
+      SELECT day, participant_id
+      FROM daily_participants_count
+      WHERE rn = 1
+    ),
+    cumulative_participants AS (
+      SELECT
+        day,
+        COUNT(participant_id) OVER (ORDER BY day) AS cumulative_participants
+      FROM first_appearance
+    )
+    SELECT
+        DISTINCT(day::TEXT),
+        cumulative_participants::INT as participants
+    FROM cumulative_participants
     WHERE day >= $1 AND day <= $2
-    GROUP BY day
     ORDER BY day
   `, [filter.from, filter.to])
   return rows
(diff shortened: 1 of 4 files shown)


stats/lib/platform-stats-fetchers.js
```diff
@@ -77,6 +77,40 @@ export const fetchDailyRewardTransfers = async (pgPool, filter) => {
...
+        day,
+        participant_id,
+        ROW_NUMBER() OVER (PARTITION BY participant_id ORDER BY day) AS rn
+      FROM daily_participants
+    ),
+    first_appearance AS (
+      SELECT day, participant_id
+      FROM daily_participants_count
+      WHERE rn = 1
+    ),
```
bajtos, line 97: It may be possible to simplify this first query as follows:

```sql
    WITH first_appearance AS (
      SELECT participant_id, MIN(day) as first_seen
      FROM daily_participants
      GROUP BY participant_id
    )
```

See https://github.com/filecoin-station/spark-stats/pull/232/files#diff-9c80f8ed9f9a7b436c7e194a16930630ea74a6af1c61484ea32e0914fe0a0c1fR177

stats/lib/platform-stats-fetchers.js
```diff
@@ -77,6 +77,40 @@ export const fetchDailyRewardTransfers = async (pgPool, filter) => {
...
+    cumulative_participants AS (
+      SELECT
+        day,
+        COUNT(participant_id) OVER (ORDER BY day) AS cumulative_participants
```
bajtos, line 101: Nice trick! It took me a bit to decipher this. The PG documentation was helpful.

https://www.postgresql.org/docs/current/tutorial-window.html
> There is another important concept associated with window functions: for each row, there is a set of rows within its partition called its window frame. Some window functions act only on the rows of the window frame, rather than of the whole partition. By default, if ORDER BY is supplied then the frame consists of all rows from the start of the partition up through the current row, plus any following rows that are equal to the current row according to the ORDER BY clause.

Having said that, I think there may be a subtle bug there. IIUC, this counts all `(day, participant_id)` pairs - i.e. if a participant was active on two days, they will contribute value `2` to the final sum.

https://www.postgresql.org/docs/current/functions-aggregate.html
> `count ( "any" ) → bigint`
> Computes the number of input rows in which the input value is not null.

Can you please check?

I would expect to see `COUNT(DISTINCT participant_id)` in the query instead. That can be slow, this StackOverflow answer suggest a faster option: https://stackoverflow.com/a/14732410/69868

Or maybe the current behaviour is what you actually want. Can you please clarify?

juliangruber: Can you verify this by looking at the tests? https://github.com/filecoin-station/spark-stats/pull/267/files#diff-bcf599781f16d5a58a3aea9fc84d1f3633de99d563fa617c59023ee1ddde4f0b Do you agree that this count is what's expected? I think so.

Tbh I haven't looked at the query in the depth that you did for the review.

juliangruber: Would you still like me to compare the current version with what's proposed, even if the output is correct? Would you like to simplify or speed up the query?

stats/test/platform-routes.test.js
```diff
@@ -362,6 +362,57 @@ describe('Platform Routes HTTP request handler', () => {
...
+
+      const res = await fetch(
+        new URL('/participants/accumulative/daily?from=2000-01-01&to=2000-01-03', baseUrl), {
+          redirect: 'manual'
+        }
+      )
+      await assertResponseStatus(res, 200)
+      const daily = await res.json()
+      assert.deepStrictEqual(daily, [
+        { day: '2000-01-01', participants: 6 },
```
bajtos, line 407: I think this test covers the question I described in my comment above.

Here is the data we have:

```js
      // 3 new participants, out of range
      await givenDailyParticipants(
        pgPools.evaluate,
        '1999-01-01',
        ['0x10', '0x20', '0x30']
      )
      // 3 new participants, 1 old participant -> 6
      await givenDailyParticipants(
        pgPools.evaluate,
        '2000-01-01',
        ['0x1', '0x2', '0x3', '0x10']
      )
```

- On 1999-01-01, three participants joined the network - '0x10', '0x20', '0x30'
- On 2000-01-01, 0x10 was active, other two existing participants were not, and three new participants joined the network - '0x1', '0x2', '0x3'

Our API endpoint gives the number of participants that joined the network before or on 2000-01-01.

The correct answer is 6 ('0x10', '0x20', '0x30', '0x1', '0x2', '0x3'), and your test asserts the correct value 👍🏻 

Then 0x4 joins on 2000-01-03 and when we ask fro the number of participants that joined the network before or on 2000-01-03, we get back 7, which is again the correct answer 👍🏻

bajtos (approved):
:shipit:

PR #268, opened 2024-12-10
Add individual daily transfers. Closes #253

### LINKED ISSUE #253: Add participant addresses in GET /transfers/daily
This removes the need to manually copy `(address, amount)` pairs into a spreadsheet

juliangruber:
Closes #253

With this change, we don't need to manually copy transfers into the rewards releases spreadsheet any more.

### DIFF
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -68,13 +68,27 @@ export const fetchParticipantsWithTopMeasurements = async (pgPool, filter) => {
  */
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
   const { rows } = await pgPool.query(`
-    SELECT day::TEXT, SUM(amount) as amount
+    SELECT day::TEXT, to_address, amount
     FROM daily_reward_transfers
     WHERE day >= $1 AND day <= $2
-    GROUP BY day
-    ORDER BY day
   `, [filter.from, filter.to])
-  return rows
+  const days = {}
+  for (const row of rows) {
+    if (!days[row.day]) {
+      days[row.day] = {
+        day: row.day,
+        amount: '0',
+        transfers: []
+      }
+    }
+    const day = days[row.day]
+    day.amount = String(BigInt(day.amount) + BigInt(row.amount))
+    day.transfers.push({
+      toAddress: row.to_address,
+      amount: row.amount
+    })
+  }
+  return Object.values(days)
 }
 
 /**
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -3,6 +3,8 @@ import { today, yesterday } from './request-helpers.js'
 
 /** @typedef {import('@filecoin-station/spark-stats-db').Queryable} Queryable */
 
+const ONE_DAY = 24 * 60 * 60 * 1000
+
 /**
  * @param {Queryable} pgPool
  * @param {import('./typings.js').DateRangeFilter} filter
@@ -67,6 +69,11 @@ export const fetchParticipantsWithTopMeasurements = async (pgPool, filter) => {
  * @param {import('./typings.js').DateRangeFilter} filter
  */
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
+  assert(
+    new Date(filter.to).getTime() - new Date(filter.from).getTime() <= 31 * ONE_DAY,
+    400,
+    'Date range must be 31 days max'
+  )
   const { rows } = await pgPool.query(`
     SELECT day::TEXT, to_address, amount
     FROM daily_reward_transfers
--- a/stats/test/platform-routes.test.js
+++ b/stats/test/platform-routes.test.js
@@ -269,6 +269,17 @@ describe('Platform Routes HTTP request handler', () => {
         }
       ])
     })
+    it('returns 400 if the date range is more than 31 days', async () => {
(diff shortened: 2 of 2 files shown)


stats/lib/platform-stats-fetchers.js
```diff
@@ -68,13 +68,27 @@ export const fetchParticipantsWithTopMeasurements = async (pgPool, filter) => {
  */
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
   const { rows } = await pgPool.query(`
-    SELECT day::TEXT, SUM(amount) as amount
+    SELECT day::TEXT, to_address, amount
```
pyropy, line 71: Would it be possible to create some kind of iterator using something like [pg-cursor](https://node-postgres.com/apis/cursor) so we could avoid loading all rows into memory in order to do calculation?

stats/lib/platform-stats-fetchers.js
```diff
@@ -68,13 +68,27 @@ export const fetchParticipantsWithTopMeasurements = async (pgPool, filter) => {
...
+  for (const row of rows) {
+    if (!days[row.day]) {
+      days[row.day] = {
+        day: row.day,
+        amount: '0',
+        transfers: []
+      }
+    }
+    const day = days[row.day]
+    day.amount = String(BigInt(day.amount) + BigInt(row.amount))
```
bajtos, line 85: Do you have any concerns about the performance impact of repeatedly converting `day.amount` between BigInt and String?

stats/lib/platform-stats-fetchers.js
bajtos: Good point.

I ran this query against the live database, and it returned less than 2000 rows. I think that's because we have data from 2024-12-02 and 2024-12-09 only. When I extrapolate that to the entire year (52 weeks) and account for double network size, we are talking about ~110k rows. Loading that into memory at once seems doable to me. WDYT?

```
EXPLAIN ANALYZE
SELECT day::TEXT, to_address, amount 
FROM daily_reward_transfers 
WHERE day >= '2024-12-01' AND day <= '2024-12-31';

                                                                        QUERY PLAN                                                                         
-----------------------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using daily_reward_transfers_day on daily_reward_transfers  (cost=0.29..72.24 rows=1862 width=87) (actual time=0.032..0.586 rows=2189 loops=1)
   Index Cond: ((day >= '2024-12-01'::date) AND (day <= '2024-12-31'::date))
 Planning Time: 0.106 ms
 Execution Time: 0.720 ms
```

Having wrote that, I have a different concern:

The execution time for two weeks' worth of data is 0.720ms. That would be like 18 seconds to query the entire year. I think we need to impose some limits on the `filter` interval this API endpoint allows.

bajtos (changes requested):
It's great that you found a way how to make this change in a backwards-compatible way!

I am concerned about the pressure this can put on our DB.

stats/lib/platform-stats-fetchers.js
pyropy: @bajtos Indeed, limiting the range would be more optimal.

juliangruber: 80cebb0

stats/lib/platform-stats-fetchers.js
juliangruber: Nope, I noticed this redundant conversion too, but kept it for simplicity.

Here's a small benchmark:

```js
> ;{ const start = new Date(); for (let i = 0; i < 1e6; i++) String(BigInt('13')); console.log(new Date() - start); }
40
```

It takes 40ms on my laptop to perform this operation one million times.

bajtos: > Nope, I noticed this redundant conversion too, but kept it for simplicity.
> 
> Here's a small benchmark:
> 
> ```js
> > ;{ const start = new Date(); for (let i = 0; i < 1e6; i++) String(BigInt('13')); console.log(new Date() - start); }
> 40
> ```
> 
> It takes 40ms on my laptop to perform this operation one million times.

I think v8 is going to optimise your inner loop into a no-op.

I wrote a benchmark that's IMO better, but it still took only like 125ms to complete, so I agree with your conclusion.

```js
const start = new Date();                                                       
let sum = '0'                                                                   
for (let i = 0n; i < 10n ** 6n; i++) {                                          
  sum = String(BigInt(sum) + i)                                                 
}                                                                               
console.log('Duration:', new Date() - start);                                   
console.log('Sum:', sum)   
```

stats/lib/platform-stats-fetchers.js
```diff
@@ -67,14 +69,33 @@ export const fetchParticipantsWithTopMeasurements = async (pgPool, filter) => {
  * @param {import('./typings.js').DateRangeFilter} filter
  */
 export const fetchDailyRewardTransfers = async (pgPool, filter) => {
+  assert(
+    new Date(filter.to).getTime() - new Date(filter.from).getTime() <= 31 * ONE_DAY,
```
bajtos, line 73: Is the code calling `fetchDailyRewardTransfers` validating that filter from/to values are valid dates?

Fortunately, invalid date strings are parsed as `NaN`, and the condition here will be false, so we are covered 👍🏻 

```
> new Date('foo').getTime() - new Date('bar').getTime() <= 310000
false
```

bajtos (approved):
:shipit:

PR #281, opened 2024-12-24
Add routes for time-to-first-byte stats

pyropy:
Adds routes for fetching global and per miner time to first byte stats.

Related to https://github.com/space-meridian/roadmap/issues/208

### DIFF
--- a/stats/lib/handler.js
+++ b/stats/lib/handler.js
@@ -16,8 +16,7 @@ import {
   fetchDailyRetrievalResultCodes,
   fetchDailyMinerRSRSummary,
   fetchDailyRetrievalTimes,
-  fetchDailyMinerRetrievalTimes,
-  fetchRetrievalTimesSummary
+  fetchDailyMinerRetrievalTimes
 } from './stats-fetchers.js'
 
 import { handlePlatformRoutes } from './platform-routes.js'
@@ -111,8 +110,6 @@ const handler = async (req, res, pgPools, SPARK_API_BASE_URL) => {
     await respond(fetchMinersRSRSummary)
   } else if (req.method === 'GET' && url === '/retrieval-result-codes/daily') {
     await respond(fetchDailyRetrievalResultCodes)
-  } else if (req.method === 'GET' && url === '/retrieval-times/summary') {
-    await respond(fetchRetrievalTimesSummary)
   } else if (req.method === 'GET' && url === '/retrieval-times/daily') {
     await respond(fetchDailyRetrievalTimes)
   } else if (req.method === 'GET' && segs[0] === 'miner' && segs[1] && segs[2] === 'retrieval-times' && segs[3] === 'summary') {
--- a/stats/lib/handler.js
+++ b/stats/lib/handler.js
@@ -15,9 +15,9 @@ import {
   fetchDealSummary,
   fetchDailyRetrievalResultCodes,
   fetchDailyMinerRSRSummary,
-  fetchDailyTTFBStats,
-  fetchDailyMinerTTFBStats,
-  fetchTTFBSummary
+  fetchDailyRetrievalTimes,
+  fetchDailyMinerRetrievalTimes,
+  fetchRetrievalTimesSummary
 } from './stats-fetchers.js'
 
 import { handlePlatformRoutes } from './platform-routes.js'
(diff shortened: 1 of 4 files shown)


stats/lib/handler.js
```diff
@@ -14,7 +14,9 @@ import {
   fetchRetrievalSuccessRate,
   fetchDealSummary,
   fetchDailyRetrievalResultCodes,
-  fetchDailyMinerRSRSummary
+  fetchDailyMinerRSRSummary,
+  fetchDailyRetrievalTimes,
```
juliangruber, line 18: "Retrieval times" isn't self explanatory - it could for example also be the timestamps at which retrievals have been attempted. What about this?

```suggestion
  fetchDailyRetrievalTimings,
```

juliangruber: See for example chrome dev tools also using "timings":

Screenshot 2025-01-05 at 22 43 59

stats/lib/stats-fetchers.js
```diff
@@ -270,3 +270,48 @@ export const fetchDailyRetrievalResultCodes = async (pgPools, filter) => {
...
+/**
+ * Fetches daily global retrieval time statistics
+ * @param {import('@filecoin-station/spark-stats-db').PgPools} pgPools
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchDailyRetrievalTimes = async (pgPools, filter) => {
+  const { rows } = await pgPools.evaluate.query(`
+    SELECT
+      day::text,
+      percentile_cont(0.5) WITHIN GROUP (ORDER BY time_to_first_byte_p50) AS ttfb_p50
```
juliangruber, line 283: Having variables `time_to_first_byte_p50` and `ttfb_p50` is confusing, as the names don't tell us how they differ. Can you think of a more semantic name for the computed column?

juliangruber (changes requested):

stats/lib/stats-fetchers.js
bajtos: There are two aspects to consider:

1. What name do we want to use in the SQL query?
2. What property name do we want to use in the response body?

Also, note that a p50 value calculated from a list of p50 values is not a true global p50 value; it's just an approximation we decided to use to keep this initial implementation simple.

In that light, I propose to use the name `ttfb` in the response.

```json
[
  { "day": "2025-01-08", "ttfb": 850 }
  { "day": "2025-01-09", "ttfb": 930 }
]
```

I think this name works well for the SQL query, too, but I don't mind using a different name there.

bajtos, line 283: `percentile_cont` returns a double precision floating number, e.g. `20.124`. This is likely to create a sense of false precision.

Let's convert the number to an integer, e.g. using the Postgres `CEIL()` function.

- We could also use `ROUND()` (10.2→10, 10.8→11), but I prefer to be conservative in what we produce and prefer to report slightly worse TTFB values rather than slightly better values.

This is a minor detail, though - I am fine with any rounding function. 

What's important to me is to produce integer values (number of milliseconds), since that's the precision we are getting from the checker ndoes.

bajtos: One more idea to consider - should we encode the unit (milliseconds) in the property name?

```json
[
  { "day": "2025-01-08", "ttfb_ms": 850 }
  { "day": "2025-01-09", "ttfb_ms": 930 }
]
```

juliangruber: `"ttfb_ms"` lgtm!

stats/lib/handler.js
bajtos: +1 to use `timings`

pyropy: Thanks for suggestion. I have renamed everything from `times` to `timings` including table, routes and functions.

stats/lib/stats-fetchers.js
pyropy: I have missed this as I was testing using round int values. I have added rounding to queries and we're now also rounding computed p50 values before inserting them to the database (in both cases we're rounding using ceil).

I am going to change tests in https://github.com/filecoin-station/spark-evaluate/pull/434 to include more odd values.

pyropy:
I am marking this ready for review, although CI will fail before https://github.com/filecoin-station/spark-evaluate/pull/434 gets merged.

stats/test/handler.test.js
```diff
@@ -767,6 +768,63 @@ describe('HTTP request handler', () => {
...
+      assert.deepStrictEqual(stats, [
+        { day: '2024-01-10', ttfb_ms: 500 },
+        { day: '2024-01-20', ttfb_ms: 1000 }
+      ])
+    })
+
+    it('lists daily retrieval timings summary for specified miner in given date range', async () => {
+      const res = await fetch(
+        new URL(
+          '/miner/f1one/retrieval-timings/summary?from=2024-01-10&to=2024-01-20',
```
bajtos, line 811: IIRC, we use `/summary` for endpoints returning aggregated data over the entire interval. This route returns data for each day within the interval. I propose to rename the route to `/miner/:id/retrieval-timings/daily`.

bajtos (approved):
LGTM 👍🏻

PR #285, opened 2025-01-09
feat: add http rsr

NikolasHaimerl:
This MR proposes the following changes:

1. Add http rsr metric
2. Change the spark-evalute dependency hash to account for the recent changes made to the spar-evalute db

Related to [Issue #192](https://github.com/space-meridian/roadmap/issues/192)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,7 @@
       }
     },
     "db/node_modules/spark-evaluate": {
-      "resolved": "git+ssh://git@github.com/filecoin-station/spark-evaluate.git#607d836f6a60b9b22c57254d6df041dcd4c1cf5f",
+      "resolved": "git+ssh://git@github.com/filecoin-station/spark-evaluate.git#3ac022aac2d5e5df7d88d33c460b45e6567d01d6",
       "dependencies": {
         "@filecoin-station/spark-impact-evaluator": "^1.2.4",
         "@glif/filecoin-address": "^3.0.12",
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@types/mocha": "^10.0.10",
         "@types/pg": "^8.11.10",
         "standard": "^17.1.2",
-        "typescript": "^5.7.2"
+        "typescript": "^5.7.3"
       }
     },
     "db": {
@@ -6758,9 +6758,10 @@
       }
     },
     "node_modules/typescript": {
-      "version": "5.7.2",
+      "version": "5.7.3",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.7.3.tgz",
+      "integrity": "sha512-84MVSjMEHP+FQRPy3pX9sTVV/INIex71s9TL2Gm5FG/WG1SqXeKyZ0k7/blY/4FdOzI12CBy1vGc4og/eus0fw==",
       "dev": true,
-      "license": "Apache-2.0",
       "bin": {
         "tsc": "bin/tsc",
         "tsserver": "bin/tsserver"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1459,9 +1459,9 @@
       }
     },
     "node_modules/@sentry/core": {
-      "version": "8.47.0",
-      "resolved": "https://registry.npmjs.org/@sentry/core/-/core-8.47.0.tgz",
-      "integrity": "sha512-iSEJZMe3DOcqBFZQAqgA3NB2lCWBc4Gv5x/SCri/TVg96wAlss4VrUunSI2Mp0J4jJ5nJcJ2ChqHSBAU48k3FA==",
+      "version": "8.48.0",
+      "resolved": "https://registry.npmjs.org/@sentry/core/-/core-8.48.0.tgz",
+      "integrity": "sha512-VGwYgTfLpvJ5LRO5A+qWo1gpo6SfqaGXL9TOzVgBucAdpzbrYHpZ87sEarDVq/4275uk1b0S293/mfsskFczyw==",
       "engines": {
         "node": ">=14.18"
       }
@@ -1511,14 +1511,6 @@
         "node": ">=14.18"
       }
     },
-    "node_modules/@sentry/node/node_modules/@sentry/core": {
-      "version": "8.48.0",
-      "resolved": "https://registry.npmjs.org/@sentry/core/-/core-8.48.0.tgz",
-      "integrity": "sha512-VGwYgTfLpvJ5LRO5A+qWo1gpo6SfqaGXL9TOzVgBucAdpzbrYHpZ87sEarDVq/4275uk1b0S293/mfsskFczyw==",
-      "engines": {
-        "node": ">=14.18"
-      }
-    },
     "node_modules/@sentry/opentelemetry": {
       "version": "8.48.0",
       "resolved": "https://registry.npmjs.org/@sentry/opentelemetry/-/opentelemetry-8.48.0.tgz",
@@ -1537,22 +1529,14 @@
         "@opentelemetry/semantic-conventions": "^1.28.0"
       }
     },
-    "node_modules/@sentry/opentelemetry/node_modules/@sentry/core": {
-      "version": "8.48.0",
-      "resolved": "https://registry.npmjs.org/@sentry/core/-/core-8.48.0.tgz",
-      "integrity": "sha512-VGwYgTfLpvJ5LRO5A+qWo1gpo6SfqaGXL9TOzVgBucAdpzbrYHpZ87sEarDVq/4275uk1b0S293/mfsskFczyw==",
-      "engines": {
-        "node": ">=14.18"
-      }
-    },
     "node_modules/@sentry/profiling-node": {
-      "version": "8.47.0",
-      "resolved": "https://registry.npmjs.org/@sentry/profiling-node/-/profiling-node-8.47.0.tgz",
-      "integrity": "sha512-YRta5UpsIOs3F/Px0XCWWybouZ62/5I+y/3Qh1u1x0PqBI42CReQTzRPGxwleXonEv5PT+P9WQoYjxRKGQCx9w==",
+      "version": "8.48.0",
+      "resolved": "https://registry.npmjs.org/@sentry/profiling-node/-/profiling-node-8.48.0.tgz",
+      "integrity": "sha512-zKUjk9647UkP3RcmupW4TT+qG8V/kXraujHaJOKnV4zDC2lcBPxswPymEveEwUlfN43UIbCEj5cDTjHWCCDxvA==",
       "hasInstallScript": true,
       "dependencies": {
-        "@sentry/core": "8.47.0",
-        "@sentry/node": "8.47.0",
+        "@sentry/core": "8.48.0",
+        "@sentry/node": "8.48.0",
         "detect-libc": "^2.0.2",
         "node-abi": "^3.61.0"
       },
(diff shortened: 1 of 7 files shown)


stats/lib/stats-fetchers.js
```diff
@@ -7,7 +7,7 @@ export const fetchRetrievalSuccessRate = async (pgPools, filter) => {
   // Fetch the "day" (DATE) as a string (TEXT) to prevent node-postgres for converting it into
   // a JavaScript Date with a timezone, as that could change the date one day forward or back.
   const { rows } = await pgPools.evaluate.query(`
-    SELECT day::text, SUM(total) as total, SUM(successful) as successful
+    SELECT day::text, SUM(total) as total, SUM(successful) as successful, SUM(successful_http) as successful_http
```
NikolasHaimerl, line 10: The database of spark-evalute leaves pas http success null. Since the query is grouped by daily retrievals I would suggest to keep the `WHERE` condition as is. Past days in which http sucess were not yet recorded will return null for the successes and success rate of http retrievals.

package-lock.json
```diff
@@ -30,7 +30,7 @@
       }
     },
     "db/node_modules/spark-evaluate": {
-      "resolved": "git+ssh://git@github.com/filecoin-station/spark-evaluate.git#607d836f6a60b9b22c57254d6df041dcd4c1cf5f",
+      "resolved": "git+ssh://git@github.com/filecoin-station/spark-evaluate.git#3ac022aac2d5e5df7d88d33c460b45e6567d01d6",
```
NikolasHaimerl, line 33: This links to the recent [commit](3ac022aac2d5e5df7d88d33c460b45e6567d01d6) that modifies the db of spark-evalute.

juliangruber:
> This MR proposes the following changes:

@NikolasHaimerl what does MR stand for?

stats/lib/stats-fetchers.js
```diff
@@ -20,7 +20,10 @@ export const fetchRetrievalSuccessRate = async (pgPools, filter) => {
     day: r.day,
     total: r.total,
     successful: r.successful,
-    success_rate: r.total > 0 ? r.successful / r.total : null
+    success_rate: r.total > 0 ? r.successful / r.total : null,
+    successful_http: r.successful_http ?? null,
+    // Successful http might be null because the column was added later
+    success_rate_http: r.total > 0 && r.successful_http ? r.successful_http / r.total : null
```
juliangruber, line 26: This has a problem: If `r.successful_http` is 0, it will be falsy, and `success_rate_http` will be `null` and not `0`. This also applies to below

juliangruber (changes requested):

stats/lib/stats-fetchers.js
juliangruber: Please also cover this edge case with a test

NikolasHaimerl:
> > This MR proposes the following changes:
> 
> @NikolasHaimerl what does MR stand for?

Just a habit from from working with Gitlab: MR= Merge Request, PR = Pull request. Means the same [thing](https://stackoverflow.com/questions/22199432/pull-request-vs-merge-request)

stats/lib/stats-fetchers.js
NikolasHaimerl: Prior to when this [PR](https://github.com/filecoin-station/spark-evaluate/pull/440) was merged there existed no `sucessful_http` entries. This means that they will be left `null` . The aggregation in this query is limited to daily values. For the days prior to the PR that added `successful_http` to the db the query will divide `null` by `total`. On the day that the PR was merged we will have slightly scewed metrics as it was merged during the day. Not all retrievals that are in `total` will have been checked for successful http retrievals. Past the day it was merged the metric is correct again. Since this is only the case for this one day in which the metric is slightly off, I thought it would be negligible. I did want to point this out though, to be sure we are aware of it.

juliangruber: Got it! I think I just didn't understand the wording with the typo. All makes sense to me 🙏

stats/lib/stats-fetchers.js
NikolasHaimerl: Thanks for catching that. I guess I missread how JS resolves bools for undefined, null and 0. 
I made the change and added a test to test all three edge cases.

stats/lib/stats-fetchers.js
```diff
@@ -20,7 +20,10 @@ export const fetchRetrievalSuccessRate = async (pgPools, filter) => {
     day: r.day,
     total: r.total,
     successful: r.successful,
-    success_rate: r.total > 0 ? r.successful / r.total : null
+    success_rate: r.total > 0 ? r.successful / r.total : null,
+    successful_http: r.successful_http ?? null,
+    // successful_http might be null because the column was added later
+    success_rate_http: r.total > 0 && !(r.successful_http === undefined || r.successful_http === null) ? r.successful_http / r.total : null
```
juliangruber, line 26:
```suggestion
    success_rate_http: r.total > 0 && typeof r.successful_http === 'number' ? r.successful_http / r.total : null
```

I think this is simpler and still covers all cases

juliangruber (changes requested):

stats/lib/stats-fetchers.js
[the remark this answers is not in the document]
juliangruber: So far we agreed not to discuss code style in PRs, and only solve this through linters. I understand these lines are getting uncomfortably long, but otoh this is a good forcing function for us to work on the proper fix for this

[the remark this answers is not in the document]
bajtos: Yeah, we haven't looked into linting SQL files yet.

I am fine with keeping the long lines if that's what others prefer.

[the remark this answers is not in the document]
juliangruber: I just don't think the PR is a good place to discuss this, let's add it to the linting issue. A issue search for "lint" in this org didn't yield any results, @bajtos do you happen to know where that issue is?

stats/lib/stats-fetchers.js
NikolasHaimerl: Not sure if this works. successful_http is a string representation of a number.
We could use something like this for the check to be more concise and reuse the expression: 
`const isValidNumber = value => value != null && !isNaN(+value)`
And then call:
`    success_rate_http: r.total > 0 && isValidNumber(r.successful_http) ? r.successful_http / r.total : null
`

stats/test/handler.test.js
```diff
@@ -75,20 +75,20 @@ describe('HTTP request handler', () => {
...
     it('returns today stats for no query string', async () => {
       const day = today()
-      await givenRetrievalStats(pgPools.evaluate, { day, total: 10, successful: 1 })
+      await givenRetrievalStats(pgPools.evaluate, { day, total: 10, successful: 1, successfulHttp: 0 })
       const res = await fetch(new URL('/retrieval-success-rate', baseUrl), { redirect: 'follow' })
       await assertResponseStatus(res, 200)
       const stats = await res.json()
       assert.deepStrictEqual(stats, [
-        { day, success_rate: 0.1, successful: '1', total: '10' }
+        { day, success_rate: 0.1, successful: '1', total: '10', successful_http: '0', success_rate_http: 0 }
```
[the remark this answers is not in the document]
NikolasHaimerl: I was using the existing formatting. 
For the function parameter for `givenRetrievalStats`  I am using pascalCase and for the sql variables I am using snake_case. 
I can't use snake_case for the function parameter because of the linting settings.

stats/lib/stats-fetchers.js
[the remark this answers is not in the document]
bajtos: https://github.com/space-meridian/roadmap/issues/166

stats/lib/stats-fetchers.js
bajtos: FYI:
node-pg represents 64-bit integers coming from Postgres as JavaScript strings to preserve precision.

Even if the column `successful_http` is represented as a 32-bit integer, `SUM(successful_http)` yields a 64-bit integer.

We could cast `successful_http` to `INT` in the Postgres query so that we receive a Number on the JavaScript side, but that would create an inconsistency with how we handle `successful`, so this is not a good solution.

I think we can trust that the database returns either `null` (when the column value is not set) or a string representation of a number.

```suggestion
    success_rate_http: r.total > 0 && r.successful_http !== null ? r.successful_http / r.total : null
```

stats/test/handler.test.js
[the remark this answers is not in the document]
bajtos: Oh, I didn't realise the first like calls `givenRetrievalStats` and this line is asserting on the data returned by the database.

In that case, you did the right thing, and this test only shows the awkwardness of our current conventions.

stats/lib/stats-fetchers.js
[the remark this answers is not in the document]
juliangruber: 🙏 https://github.com/space-meridian/roadmap/issues/166#issuecomment-2580516671

stats/lib/stats-fetchers.js
NikolasHaimerl: Done!

stats/lib/stats-fetchers.js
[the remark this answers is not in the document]
NikolasHaimerl: Done!

stats/lib/stats-fetchers.js
```diff
@@ -207,7 +210,7 @@ export const fetchParticipantRewardTransfers = async (pgPools, { from, to }, add
  */
 export const fetchMinersRSRSummary = async (pgPools, filter) => {
   const { rows } = await pgPools.evaluate.query(`
-    SELECT miner_id, SUM(total) as total, SUM(successful) as successful
+    SELECT miner_id, SUM(total) as total, SUM(successful) as successful, SUM(successful_http) as successful_http
```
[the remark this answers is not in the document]
NikolasHaimerl: Done!

stats/lib/stats-fetchers.js
```diff
@@ -232,7 +238,7 @@ export const fetchMinersRSRSummary = async (pgPools, filter) => {
  */
 export const fetchDailyMinerRSRSummary = async (pgPools, { from, to }, minerId) => {
   const { rows } = await pgPools.evaluate.query(`
-    SELECT day::TEXT, SUM(total) as total, SUM(successful) as successful
+    SELECT day::TEXT, SUM(total) as total, SUM(successful) as successful, SUM(successful_http) as successful_http
```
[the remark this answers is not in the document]
NikolasHaimerl: Done

stats/test/handler.test.js
```diff
@@ -208,19 +208,40 @@ describe('HTTP request handler', () => {
...
+      await givenRetrievalStats(pgPools.evaluate, { day, total: 10, successful: 0, minerId: 'f1two', successfulHttp: 0 })
       const res = await fetch(new URL('/retrieval-success-rate?nonZero=true', baseUrl), { redirect: 'follow' })
       await assertResponseStatus(res, 200)
       const stats = await res.json()
       assert.deepStrictEqual(stats, [
-        { day, success_rate: 0.1, successful: '1', total: '10' }
+        { day, success_rate: 0.1, successful: '1', total: '10', successful_http: '1', success_rate_http: 0.1 }
+      ])
+    })
+    it('handles successful_http values 0, null, undefined', async () => {
```
bajtos, line 226: Please add a similar test for per-miner RSR endpoint. I find it important to verify that `SUM()` and our logic detecting `null` values work as expected.

bajtos (commented):
I realised we are missing one test (see my comment below). The rest LGTM.

stats/test/handler.test.js
NikolasHaimerl: Done!

stats/test/handler.test.js
```diff
@@ -237,12 +239,43 @@ describe('HTTP request handler', () => {
...
+      let stats = await res.json()
       assert.deepStrictEqual(stats, [
         { day: '2024-01-20', success_rate: 0.1, successful: '1', total: '10', successful_http: '0', success_rate_http: 0 },
         { day: '2024-01-21', success_rate: 0.1, successful: '1', total: '10', successful_http: null, success_rate_http: null },
         { day: '2024-01-22', success_rate: 0.1, successful: '1', total: '10', successful_http: null, success_rate_http: null }
       ])
+
+      res = await fetch(
+        new URL(
+          '/miners/retrieval-success-rate/summary?from=2024-01-20&to=2024-01-22',
```
bajtos, line 251: This test case (`it('handles successful_http values 0, null, undefined',`) is nested inside the test suite `describe('GET /retrieval-success-rate',`. It should test the endpoint `GET /retrieval-success-rate` only.

Please move this new block testing `/miners/retrieval-success-rate/summary` to a new test case inside the test suite `describe('GET /miners/retrieval-success-rate/summary'` later in this file.

https://github.com/filecoin-station/spark-stats/blob/f4918ee55e46138df3824a250d3a04ae100e6b2b/stats/test/handler.test.js#L463

bajtos (approved):
LGTM 👏🏻 

Since @juliangruber requested some changes in the past, please give him ~24 hours to review the final version before you proceed to land this pull request.

juliangruber (approved):
👏

PR #293, opened 2025-01-21
stats: add `fastify`

juliangruber:


stats/lib/app.js
```diff
@@ -0,0 +1,151 @@
...
+    origin: ['http://localhost:3000', 'app://-']
+  })
+  app.register(urlData)
+
+  app.get('/deals/daily', async (/** @type {RequestWithFilter} */ request, reply) => {
+    await withFilter(request, reply, filter => {
+      return fetchDailyDealStats(pgPools, filter)
+    })
+  })
+  app.get('/deals/summary', async (/** @type {RequestWithFilter} */ request, reply) => {
```
pyropy, line 58: I feel like we could maybe introduce new helper for registering a handler as most of the handlers use `withFilter` helper anyways.

stats/lib/app.js
```diff
@@ -0,0 +1,151 @@
...
+}) => {
+  const app = Fastify({ logger })
+  Sentry.setupFastifyErrorHandler(app)
+
+  app.register(cors, {
+    origin: ['http://localhost:3000', 'app://-']
+  })
+  app.register(urlData)
+
+  app.get('/deals/daily', async (/** @type {RequestWithFilter} */ request, reply) => {
```
pyropy, line 53: What do you think about introducing `handlers.js` back or extract registering handlers into a separate function and leaving `createApp` function only for creating and configuring the `Fastify` app object and registering middlewares, etc?

stats/lib/app.js
```diff
@@ -0,0 +1,151 @@
...
+
+  return app
+}
+
+/**
+ * @param {Fastify.FastifyRequest} request
+ * @param {Fastify.FastifyReply} reply
+ * @param {string} SPARK_API_BASE_URL
+ */
+const redirectToSparkApi = (request, reply, SPARK_API_BASE_URL) => {
```
pyropy, line 145: It's a nitpick but I feel like this name could be more generic, maybe something like `redirectToURL` or something like that.

stats/lib/request-helpers.js
```diff
@@ -18,22 +18,17 @@ export const yesterday = () => getLocalDayAsISOString(new Date(Date.now() - 24 *
...
- * @param {import('node:http').ServerResponse} res
- * @param {PgPools} pgPools
- * @param {(pgPools: PgPools, filter: FilterType, pathParams: object) => Promise<object[]>} fetchStatsFn
+ * @param {RequestWithFilter} request
+ * @param {import('fastify').FastifyReply} reply
+ * @param {(filter: FilterType) => Promise<object[]>} fn
  */
-export const getStatsWithFilterAndCaching = async (pathname, pathParams, searchParams, res, pgPools, fetchStatsFn) => {
-  const filter = Object.fromEntries(searchParams)
+export const withFilter = async (request, reply, fn) => {
```
pyropy, line 25: I like this name you've introduced much better, good job 👍🏻

pyropy (commented):
I would break code apart a bit more, maybe introduce some helper functions to reduce the repetitive code. Otherwise I like that we're moving away from the built-in http package.

stats/lib/app.js
juliangruber: Did that in https://github.com/filecoin-station/spark-stats/pull/293/commits/abff832be3ab4c4d8e1326ae48276870b540aec4 :)

stats/lib/app.js
juliangruber: What would be the advantage of that?

stats/lib/app.js
juliangruber: But it's not that. A generic redirect would be `reply.redirect(url, status)`. This is a specific redirect to spark-api

stats/lib/app.js
pyropy: Nice, that looks much cleaner than the thing we had before, good job!

stats/lib/app.js
pyropy: In that case `createApp` function would do only only thing -- creating the app. Also in my opinion code would be much cleaner. Plus we already are doing that for platform routes so it would make sense to do it for stat routes.

stats/lib/app.js
pyropy: Fair enough, let's leave it as `redirectToSparkApi` then.

stats/lib/app.js
```diff
@@ -0,0 +1,129 @@
...
+    app.get('/participants/daily', async (/** @type {RequestWithFilter} */ request, reply) => {
+      reply.send(await fetchDailyParticipants(pgPools, request.filter))
+    })
+    app.get('/participants/monthly', async (/** @type {RequestWithFilter} */ request, reply) => {
+      reply.send(await fetchMonthlyParticipants(pgPools, request.filter))
+    })
+    app.get('/participants/change-rates', async (/** @type {RequestWithFilter} */ request, reply) => {
+      reply.send(await fetchParticipantChangeRates(pgPools, request.filter))
+    })
+    app.get('/participant/:address/scheduled-rewards', async (/** @type {RequestWithFilterAndAddress} */ request, reply) => {
```
juliangruber, line 75: this type is awkward

stats/lib/app.js
juliangruber: It would do more than creating the app though. It would configure the app, and add plugins. To me, the app definition, with all routes that make sense, should be in one file. It also means you have to jump to less files to get to the often important part.

I want to give your suggestion a try however, and implemented it in 98c04bd. What do you think?

stats/bin/spark-stats.js
```diff
@@ -12,15 +10,14 @@ const {
...
+  logger: {
+    level: ['1', 'true'].includes(REQUEST_LOGGING) ? 'info' : 'error'
+  }
+})
 console.log('Starting the http server on host %j port %s', HOST, PORT)
-server.listen(Number(PORT), HOST)
-await once(server, 'listening')
-console.log(`http://${HOST}:${PORT}`)
+const baseUrl = app.listen({ port: Number(PORT), host: HOST })
+console.log(baseUrl)
```
NikolasHaimerl, line 23:
```suggestion
console.log(`Server is listening on ${baseUrl}`)
```

pyropy (approved):

stats/lib/app.js
juliangruber: ping @pyropy

bajtos (approved):
Nice! 🤩 

I am not able to fully assess all potential ramifications of these changes, but I also don't see any obvious problems. I think it's best to work on this incrementally, land what you have here and then iteratively improve based on what we learn while implementing future changes.

:shipit:

stats/bin/spark-stats.js
juliangruber: This is a change in behavior, it should be clear what the address being logged is

PR #311, opened 2025-02-04
Observe daily desktop users count

pyropy:
Adds loop that observes daily desktop user count and saves it to the database.

Closes https://github.com/CheckerNetwork/roadmap/issues/207


observer/lib/observer.js
```diff
@@ -129,3 +129,32 @@ export const observeRetrievalResultCodes = async (pgPoolStats, influxQueryApi) =
     ])
   }
 }
+
+export const observeDesktopUsers = async (pgPoolStats, influxQueryApi) => {
+  const rows = await influxQueryApi.collectRows(`
+    from(bucket: "station-machines")
```
NikolasHaimerl, line 135: Is there also a way for us to test this query?

NikolasHaimerl (changes requested):

observer/lib/observer.js
pyropy: Good question, I think only way to test the query itself would be to spin up InfluxDB instance and execute the query against it.

db/migrations/006.do.daily-desktop-users.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE daily_desktop_users (
+    day DATE NOT NULL,
+    platform TEXT NOT NULL,
+    user_count INT NOT NULL,
```
juliangruber, line 4: INT has a max value of 2,147,483,647. I think that should be fine for desktop user count for the foreseeable future

observer/lib/observer.js
```diff
@@ -129,3 +129,32 @@ export const observeRetrievalResultCodes = async (pgPoolStats, influxQueryApi) =
...
+      |> range(start: -24h)
```
juliangruber, line 136: Do you have any concerns about this query being defined as over the last 24h? This means depending on when this function is run, you will get a different count for the day. It would be safer to query the full 24h from the last day, always, for reproducible results.

pyropy: Good point, I think it would make sense to query results for previous day with exact start and end time.

pyropy: Found a really nice, clean way to do it using [boundaries.yesterday](https://docs.influxdata.com/flux/v0/stdlib/experimental/date/boundaries/yesterday/) function, only problem is that it's part of the experimental package. What do you think, should we use that or have it passed down from javascript?

observer/lib/observer.js
pyropy: I think adding new function to [dry-run.js](https://github.com/CheckerNetwork/spark-stats/blob/main/observer/bin/dry-run.js) would be sufficient , similar to how @bajtos has done it in his [PR](https://github.com/CheckerNetwork/spark-stats/pull/319), as we would avoid introducing too much new changes in this PR. However we can make a follow-up issue to introduce tests for InfluxDB queries.

juliangruber (changes requested):

observer/lib/observer.js
```diff
@@ -129,3 +129,30 @@ export const observeRetrievalResultCodes = async (pgPoolStats, influxQueryApi) =
     ])
   }
 }
+
+export const observeYesterdayDesktopUsers = async (pgPoolStats, influxQueryApi) => {
+  const rows = await influxQueryApi.collectRows(`
+    import "experimental/date/boundaries"
+    yesterday = boundaries.yesterday()
```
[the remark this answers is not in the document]
pyropy: We're using Influx Cloud which has Flux version `0.194.3` and `boundaries.yesterday` has been marked as experimental since version `0.177.1`. Sadly I think we'll have to resort do getting these values inside the JavaScript.

Screenshot 2025-02-13 at 10 19 52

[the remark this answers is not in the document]
juliangruber: Oh wow I misread, It was stable once and then marked as experimental later :O

observer/lib/observer.js
pyropy: I have updated `INFLUXDB_TOKEN` to have read permissions to `station-machines` bucket and I have added new function to dry-run.

observer/lib/observer.js
```diff
@@ -130,3 +130,54 @@ export const observeRetrievalResultCodes = async (pgPoolStats, influxQueryApi) =
     rows.map(r => r._value)
   ])
 }
+
+export const observeYesterdayDesktopUsers = async (pgPoolStats, influxQueryApi) => {
+  // TODO: Replace with Flux boundaries.yesterday() once it becomes part of stable API
+  const yesterday = getYesterdayBoundaries()
```
juliangruber, line 136: Great to abstract this into a function 👏

observer/lib/observer.js
```diff
@@ -130,3 +130,54 @@ export const observeRetrievalResultCodes = async (pgPoolStats, influxQueryApi) =
...
+
+  // Create start of yesterday
+  const start = new Date(now)
+  start.setDate(start.getDate() - 1) // Move to yesterday
+  start.setUTCHours(0, 0, 0, 0) // Set to start of day
+
+  // Create end of yesterday
+  const stop = new Date(now)
+  stop.setDate(stop.getDate() - 1) // Move to yesterday
+  stop.setUTCHours(23, 59, 59, 999) // Set to end of day
```
juliangruber, line 177:
```suggestion
  stop.setUTCHours(0, 0, 0, 0) // Set to end of day
```

range stop is exclusive, see https://docs.influxdata.com/flux/v0/stdlib/universe/range/

juliangruber (changes requested):

observer/lib/telemetry.js
```diff
@@ -6,7 +6,7 @@ const debug = createDebug('spark:observer:telemetry')
 export const createInflux = token => {
   const influx = new InfluxDB({
     url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
-    // bucket permissions: spark-evaluate:read spark-observer:write
+    // bucket permissions: spark-evaluate:read spark-observer:write station-machines:read
```
juliangruber, line 9: did you update the token?

juliangruber (approved):

observer/lib/telemetry.js
pyropy: I did in the github secrets, I've also added the token to fly.io now as

NikolasHaimerl (approved):

PR #327, opened 2025-02-18
Add daily desktop users route

pyropy:
Adds route that returns total number of desktop users per day with platform breakdown.

Related to:
-  https://github.com/CheckerNetwork/roadmap/issues/207
- #311

PR #329, opened 2025-02-18
Remove platform column and add endpoint to read daily desktop users

pyropy:
### Motivation

In my 1:1 with @patrickwoodhead, I raised the question of whether we need to track the platform used by desktop users to run the app. We concluded that we're only interested in the number of users running the Station app.

### Changelog

- Drop `platform` column from `daily_desktop_users` table
- Simplify InfluxDB query to only return daily desktop users count
- Add endpoint to read daily desktop users count in a given time range

Follow up #311 
Related to https://github.com/CheckerNetwork/roadmap/issues/207

### DIFF
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -164,3 +164,20 @@ export const fetchParticipantsSummary = async (pgPool) => {
     participant_count: Number(rows[0].count)
   }
 }
+
+/**
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchDailyDesktopUsers = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT
+      day::TEXT,
+      user_count::INT
+    FROM daily_desktop_users
+    WHERE day >= $1 AND day <= $2
+    ORDER BY day`,
+  [filter.from, filter.to])
+
+  return rows
+}
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -173,7 +173,7 @@ export const fetchDailyDesktopUsers = async (pgPool, filter) => {
   const { rows } = await pgPool.query(`
     SELECT
       day::TEXT,
-      user_count::INT
+      user_count
     FROM daily_desktop_users
     WHERE day >= $1 AND day <= $2
     ORDER BY day`,
--- a/db/migrations/007.do.drop-platform-column-from-daily-desktop-users.sql
+++ b/db/migrations/007.do.drop-platform-column-from-daily-desktop-users.sql
@@ -1,0 +1,3 @@
+ALTER TABLE daily_desktop_users DROP CONSTRAINT daily_desktop_users_pkey;
+ALTER TABLE daily_desktop_users DROP COLUMN platform;
+ALTER TABLE daily_desktop_users ADD PRIMARY KEY (day);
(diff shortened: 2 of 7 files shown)


bajtos (changes requested):
I want to keep the data about platform-based usage, to allow us to make data-based decision about how much effort to invest into which platform.

Let's keep the platform data in the database & REST API response and hide it at the presentation layer.

Alternatively, keep the platform data in the database and modify the REST API response to return aggregated data across all platforms.

bajtos:
@pyropy IIUC, this is a follow-up for https://github.com/CheckerNetwork/spark-stats/pull/311 and https://github.com/CheckerNetwork/roadmap/issues/207, is that correct? Please cross-reference pull requests and issues in the future, e.g. in the PR/issue description. It makes it easier to find the relevant context when looking at historical work.

bajtos:
Please add the spark-stats API endpoint exposing the new time series data - include it in this PR.

stats/lib/platform-stats-fetchers.js
```diff
@@ -164,3 +164,20 @@ export const fetchParticipantsSummary = async (pgPool) => {
...
+
+/**
+ * @param {Queryable} pgPool
+ * @param {import('./typings.js').DateRangeFilter} filter
+ */
+export const fetchDailyDesktopUsers = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT
+      day::TEXT,
+      user_count::INT
```
juliangruber, line 176: For my understanding, why are we casting to int here?

juliangruber (changes requested):
@pyropy please update the PR title to match the new intend

stats/lib/platform-stats-fetchers.js
pyropy: In the query that was part of #327 I had to explicitly cast to int. I've just removed it here and it works fine. Thanks for raising this up 👍🏻

stats/lib/platform-stats-fetchers.js
```diff
@@ -164,3 +164,20 @@ export const fetchParticipantsSummary = async (pgPool) => {
...
+ */
+export const fetchDailyDesktopUsers = async (pgPool, filter) => {
+  const { rows } = await pgPool.query(`
+    SELECT
+      day::TEXT,
+      user_count
+    FROM daily_desktop_users
+    WHERE day >= $1 AND day <= $2
+    ORDER BY day`,
+  [filter.from, filter.to])
```
NikolasHaimerl, line 180: Should we assert that from and to are actual dates? 
I think the types are not checked during runtime so as a user I could send and string inside of an object `Filter {from:str1, to:str2}`. Correct me if I am wrong here?

juliangruber (approved):

stats/lib/platform-stats-fetchers.js
pyropy: Yes both `from` and `to` are strings and are not converted to dates. They are validated inside the pre-handler hook. Inside the hook we enforce ISO date format:

```javascript
  // Trim time from date-time values that are typically provided by Grafana

  const matchFrom = filter.from.match(/^(\d{4}-\d{2}-\d{2})(T\d{2}:\d{2}:\d{2}\.\d{3}Z)?$/)
  assert(matchFrom, 400, '"from" must have format YYYY-MM-DD or YYYY-MM-DDThh:mm:ss.sssZ')
  if (matchFrom[2]) {
    filter.from = matchFrom[1]
    shouldRedirect = true
  }

  const matchTo = filter.to.match(/^(\d{4}-\d{2}-\d{2})(T\d{2}:\d{2}:\d{2}\.\d{3}Z)?$/)
  assert(matchTo, 400, '"to" must have format YYYY-MM-DD or YYYY-MM-DDThh:mm:ss.sssZ')
  if (matchTo[2]) {
    filter.to = matchTo[1]
    shouldRedirect = true
  }
```

We could potentially try to convert it to Date object but I think there's no need for that currently.

PR #330, opened 2025-02-25
feat: per client rsr api endpoint

NikolasHaimerl:
### CHANGELOG

1. Per Client RSR summary endpoint
2. Clients RSR summary endpoint

The two new endpoints closely follow the implementation of the [per miner](https://github.com/CheckerNetwork/spark-stats/blob/main/stats/lib/routes.js#L67) and [miners](https://github.com/CheckerNetwork/spark-stats/blob/main/stats/lib/routes.js#L52) RSR summary endpoints. These existing endpoints expose `retrieval_stats` for miner-related statistics.  

This PR introduces the same logic for client-related statistics by exposing the `daily_client_retrieval_stats` table, which was created in this recent [PR](https://github.com/CheckerNetwork/spark-evaluate/pull/481/files).  

Additionally, the tests for these new endpoints follow the same pattern as those for the existing miner RSR stats endpoints, ensuring consistency and reliability.

Closes subtask of https://github.com/CheckerNetwork/roadmap/issues/193

### DIFF
--- a/stats/lib/stats-fetchers.js
+++ b/stats/lib/stats-fetchers.js
@@ -1,3 +1,4 @@
+import assert from 'node:assert'
 /** @typedef {import('@filecoin-station/spark-stats-db').PgPools} PgPools */
 /**
  * @param {PgPools} pgPools
@@ -376,15 +377,21 @@ export const fetchClientsRSRSummary = async (pgPools, filter) => {
     filter.from,
     filter.to
   ])
-  const stats = rows.map(r => ({
-    client_id: r.client_id,
-    total: r.total,
-    successful: r.successful,
-    success_rate: r.total > 0 ? r.successful / r.total : null,
-    successful_http: r.successful_http ?? null,
-    // successful_http might be null because the column was added later
-    success_rate_http: r.total > 0 && r.successful_http !== null ? r.successful_http / r.total : null
-  }))
+  const stats = rows.map(r => {
+    // All values are configured to be NOT NULL on the database level
+    assert.ok(r.successful !== null, 'successful should not be null')
+    assert.ok(r.total !== null, 'total should not be null')
+    assert.ok(r.successful_http !== null, 'successful_http should not be null')
+    return {
+      client_id: r.client_id,
+      total: r.total,
+      successful: r.successful,
+      success_rate: r.total > 0 ? r.successful / r.total : null,
+      successful_http: r.successful_http,
+      // successful_http might be null because the column was added later
+      success_rate_http: r.total > 0 ? r.successful_http / r.total : null
+    }
+  })
   return stats
 }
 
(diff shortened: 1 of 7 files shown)


NikolasHaimerl:
Waiting for release of new [npm package](https://www.npmjs.com/package/@filecoin-station/spark-evaluate) for `spark-evaluate`. This is required for the use of the new database which exposes daily per client retrieval stats.

stats/lib/stats-fetchers.js
```diff
@@ -356,3 +357,77 @@ export const fetchMinersTimingsSummary = async (pgPools, { from, to }) => {
...
+    GROUP BY client_id
+   `, [
+    filter.from,
+    filter.to
+  ])
+  const stats = rows.map(r => {
+    // All values are configured to be NOT NULL on the database level
+    assert.ok(r.successful !== null, 'successful should not be null')
+    assert.ok(r.total !== null, 'total should not be null')
+    assert.ok(r.successful_http !== null, 'successful_http should not be null')
```
juliangruber, line 384: If they are configured to be not null, we don't need to assert the values, right?

juliangruber: same applies below

pyropy (commented):
Everything looks really good 👏🏻 Let's see if we do actually need those asserts before we merge this PR.

stats/lib/stats-fetchers.js
pyropy: I had the same comment but for some reason it was not published. 

I was referring back to a abbreviation Miro has linked previously in his comments to PRs -- [YANGI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). If we don't actually need it, let's delete it.

NikolasHaimerl: My thinking was that if we make a change to the database specifications (which is not in this repo), this would act as a safeguard to prevent unintentional arithmetic operations with null. And it would give us a better error message if this scenario should occur.

juliangruber: I understand the motivation - we're codifying our assumptions about how the database should behave. I think this behavior makes sense if we can't control the dependency we're working with - eg if it's an external API. Here the situation is different: We control both the dependency and all dependents. It's our task to keep them in sync, and if we change the schema we need to check all places where the schema is used. This is a downside of the current system, and I would like to see us move away from shared databases, and switch to well defined APIs instead. But for now, we are safe to rely on the schema behaving as intended.

NikolasHaimerl: Fair Point. I'll remove the assertions then.

juliangruber (approved):

stats/lib/stats-fetchers.js
NikolasHaimerl: Adressed in abed651

PR #334, opened 2025-02-26
fix: InfluxDB bucket permission comment

pyropy:
- Fix InfluxDB bucket permission comment to reflect correct permissions for InfluxDB token

### DIFF
--- a/observer/lib/telemetry.js
+++ b/observer/lib/telemetry.js
@@ -6,7 +6,7 @@ const debug = createDebug('spark:observer:telemetry')
 export const createInflux = token => {
   const influx = new InfluxDB({
     url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
-    // bucket permissions: spark-evaluate:read spark-observer:write station-machines:read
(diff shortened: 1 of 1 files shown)


observer/lib/telemetry.js
```diff
@@ -6,7 +6,7 @@ const debug = createDebug('spark:observer:telemetry')
 export const createInflux = token => {
   const influx = new InfluxDB({
     url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
-    // bucket permissions: spark-evaluate:read spark-observer:write station-machines:read
+    // bucket permissions: spark-evaluate:read spark-observer:write station:read
```
juliangruber, line 9: Does this mirror the permissions the actual token has?

pyropy: Yes, are currently using token with those permissions. 

In the #311 we've introduced query that queries `station-machines` bucket. In the follow-up PR #328 we've changed the bucket we query but this comment was not updated.

juliangruber (approved):

PR #340, opened 2025-03-09
Optimize DB Schema & Query for Top-Earning Leaderboard

Hany-Almnaem:
Refactor DB schema: Add participant_id & foreign keys for performance

Changes Made:

- 	Added address_mapping table and migrations.
- 	Updated daily_scheduled_rewards and daily_reward_transfers to reference participant_id.
- 	New approach with stable ID-based approach.

Performance:
	
EXPLAIN ANALYZE

Migration Steps:
                new migration files/
                ├── ...
                ├── 008.do.create-address-mapping-table.sql
                ├── 009.do.backfill-address-mapping.sql
                ├── 010.do.add-participant_id-columns.sql
                ├── 011.do.populate-participant-ids.sql
                ├── 012.do.add-foreign-keys-and-indexes.sql
  
  Please review and let me know if any further changes are needed.

Close [CheckerNetwork/roadmap#178](https://github.com/CheckerNetwork/roadmap/issues/178)

### DIFF
--- a/.vscode/easycode.ignore
+++ b/.vscode/easycode.ignore
@@ -1,0 +1,12 @@
+node_modules/
+dist/
+vendor/
+cache/
+.*/
+*.min.*
+*.test.*
+*.spec.*
+*.bundle.*
+*.bundle-min.*
+*.log
+package-lock.json
--- a/db/migrations/008.do.create-address-mapping.sql
+++ b/db/migrations/008.do.create-address-mapping.sql
@@ -1,0 +1,5 @@
+-- Create a stable address mapping table
+CREATE TABLE address_mapping (
+    participant_id SERIAL PRIMARY KEY,            -- Auto-incremented stable ID
+    participant_address TEXT UNIQUE NOT NULL      -- Ensures each address has only one ID
+);
--- a/db/migrations/009.do.backfill-address-mapping.sql
+++ b/db/migrations/009.do.backfill-address-mapping.sql
@@ -1,0 +1,4 @@
+-- Migration to backfill existing participant addresses into address_mapping table
+INSERT INTO address_mapping (participant_address)
+SELECT DISTINCT participant_address FROM daily_scheduled_rewards
+ON CONFLICT (participant_address) DO NOTHING;
--- a/db/migrations/012.do.add-foreign-keys-and-indexes.sql
+++ b/db/migrations/012.do.add-foreign-keys-and-indexes.sql
@@ -1,0 +1,10 @@
+-- Foreign Keys
+ALTER TABLE daily_scheduled_rewards
+ADD FOREIGN KEY (participant_id) REFERENCES address_mapping(participant_id);
+
+ALTER TABLE daily_reward_transfers
+ADD FOREIGN KEY (participant_id) REFERENCES address_mapping(participant_id);
+
+-- Create indexes for better performance
+CREATE INDEX idx_daily_scheduled_rewards_pid_day ON daily_scheduled_rewards (participant_id, day DESC);
+CREATE INDEX idx_daily_reward_transfers_pid_day ON daily_reward_transfers (participant_id, day);
--- a/stats/lib/platform-stats-fetchers.js
+++ b/stats/lib/platform-stats-fetchers.js
@@ -136,22 +136,27 @@ export const fetchTopEarningParticipants = async (pgPool, filter) => {
   assert(filter.to === today(), 400, 'filter.to must be today, other values are not supported')
   const { rows } = await pgPool.query(`
     WITH latest_scheduled_rewards AS (
-      SELECT DISTINCT ON (participant_address) participant_address, scheduled_rewards
-      FROM daily_scheduled_rewards
-      ORDER BY participant_address, day DESC
+      SELECT dsr.participant_id, dsr.scheduled_rewards
+      FROM daily_scheduled_rewards dsr
+      WHERE dsr.day = (
+        SELECT MAX(day) 
+        FROM daily_scheduled_rewards dsr2 
+        WHERE dsr2.participant_id = dsr.participant_id
+      )
     )
     SELECT
-      COALESCE(drt.to_address, lsr.participant_address) as participant_address,
-      COALESCE(SUM(drt.amount), 0) + COALESCE(lsr.scheduled_rewards, 0) as total_rewards
+      COALESCE(drt.participant_id, lsr.participant_id) AS participant_id,
+      COALESCE(SUM(drt.amount), 0) + COALESCE(lsr.scheduled_rewards, 0) AS total_rewards
     FROM daily_reward_transfers drt
     FULL OUTER JOIN latest_scheduled_rewards lsr
-      ON drt.to_address = lsr.participant_address
+      ON drt.participant_id = lsr.participant_id
     WHERE (drt.day >= $1 AND drt.day <= $2) OR drt.day IS NULL
-    GROUP BY COALESCE(drt.to_address, lsr.participant_address), lsr.scheduled_rewards
+    GROUP BY COALESCE(drt.participant_id, lsr.participant_id), lsr.scheduled_rewards
     ORDER BY total_rewards DESC
-  `, [filter.from, filter.to])
-  return rows
-}
+  `, [filter.from, filter.to]);
+
+  return rows;
+};
 
 /**
  * @param {Queryable} pgPool
--- a/db/migrations/010.do.add-participant_id-columns.sql
+++ b/db/migrations/010.do.add-participant_id-columns.sql
@@ -1,0 +1,2 @@
+ALTER TABLE daily_scheduled_rewards ADD COLUMN participant_id INTEGER;
+ALTER TABLE daily_reward_transfers ADD COLUMN participant_id INTEGER;
--- a/db/migrations/011.do.populate-participant_id-columns.sql
+++ b/db/migrations/011.do.populate-participant_id-columns.sql
@@ -1,0 +1,11 @@
+-- Update daily_scheduled_rewards
+UPDATE daily_scheduled_rewards dsr
+SET participant_id = am.participant_id
+FROM address_mapping am
+WHERE dsr.participant_address = am.participant_address;
+
+-- Update daily_reward_transfers
+UPDATE daily_reward_transfers drt
+SET participant_id = am.participant_id
+FROM address_mapping am
+WHERE drt.to_address = am.participant_address;


.vscode/easycode.ignore
```diff
@@ -0,0 +1,12 @@
+node_modules/
+dist/
+vendor/
+cache/
+.*/
+*.min.*
+*.test.*
```
pyropy, line 7: Could you please remove this file 🙏🏻

pyropy (changes requested):
Great work, @Hany-Almnaem ! 🚀 Thanks for your submission! 🙏🏻

I'm wondering if we should reduce the number of migrations by combining some of them into a single file. For example, we could merge new table creation with backfilling or group the changes to daily_scheduled_rewards and daily_reward_transfers tables along with their backfilling. What do you think?

Let's also wait for feedback from others, but overall, this is looking great! 👍🏻

Hany-Almnaem:
@pyropy Thanks for your feedback!
> I'm wondering if we should reduce the number of migrations by combining some of them into a single file.

It's a great Idea to reduce the number of migrations. but I recommend keeping each logical change in its own migration – it’s usually easier to debug and revert that way. However, combining steps is possible if it won’t cause issues later. In general, small, incremental migrations are the safest bet.

bajtos:
@Hany-Almnaem thank you for the pull request. To clarify: is this superseding your earlier PR #324? Can we close #324 now?

There are several failed CI checks, please take a look and fix them. (You should be able to reproduce them locally by running `npm run test`.)

db/migrations/008.do.create-address-mapping.sql
```diff
@@ -0,0 +1,5 @@
+-- Create a stable address mapping table
+CREATE TABLE address_mapping (
+    participant_id SERIAL PRIMARY KEY,            -- Auto-incremented stable ID
+    participant_address TEXT UNIQUE NOT NULL      -- Ensures each address has only one ID
+);
```
bajtos, line 5: We will be frequently looking up participants using `participant_address`. It's important to add an index for that. (This is a common mistake that I make myself, too.)

db/migrations/008.do.create-address-mapping.sql
```diff
@@ -0,0 +1,5 @@
+-- Create a stable address mapping table
+CREATE TABLE address_mapping (
```
bajtos, line 2: Let's use the same table definition as we have in spark-evaluate for consistency.

https://github.com/CheckerNetwork/spark-evaluate/blob/36629d4f41cf84afc04c23fb41c41aa98c13e583/migrations/003.do.daily-participants.sql#L1-L4
(We can leave out the NOT NULL part as it's implied by PRIMARY KEY.)

```sql
CREATE TABLE participants (
  id SERIAL PRIMARY KEY,
  participant_address TEXT NOT NULL UNIQUE
);
```

db/migrations/009.do.backfill-address-mapping.sql
```diff
@@ -0,0 +1,4 @@
+-- Migration to backfill existing participant addresses into address_mapping table
+INSERT INTO address_mapping (participant_address)
+SELECT DISTINCT participant_address FROM daily_scheduled_rewards
+ON CONFLICT (participant_address) DO NOTHING;
```
bajtos, line 4: How about the participants recorded in `daily_reward_transfers`? I don't remember enough details to confidently tell whether every participant address found in `daily_reward_transfers` exists in `daily_scheduled_rewards ` too.

Have you considered running a similar INSERT query but SELECTing from `daily_reward_transfers`?

I propose keeping both queries in this one file for simplicity.

db/migrations/012.do.add-foreign-keys-and-indexes.sql
```diff
@@ -0,0 +1,10 @@
+-- Foreign Keys
+ALTER TABLE daily_scheduled_rewards
+ADD FOREIGN KEY (participant_id) REFERENCES address_mapping(participant_id);
+
+ALTER TABLE daily_reward_transfers
+ADD FOREIGN KEY (participant_id) REFERENCES address_mapping(participant_id);
+
+-- Create indexes for better performance
+CREATE INDEX idx_daily_scheduled_rewards_pid_day ON daily_scheduled_rewards (participant_id, day DESC);
+CREATE INDEX idx_daily_reward_transfers_pid_day ON daily_reward_transfers (participant_id, day);
```
bajtos, line 10: Let's also drop the old indices referencing `participant_address` instead of `participant_id`.

stats/lib/platform-stats-fetchers.js
```diff
@@ -136,22 +136,27 @@ export const fetchTopEarningParticipants = async (pgPool, filter) => {
...
-      SELECT DISTINCT ON (participant_address) participant_address, scheduled_rewards
-      FROM daily_scheduled_rewards
-      ORDER BY participant_address, day DESC
+      SELECT dsr.participant_id, dsr.scheduled_rewards
+      FROM daily_scheduled_rewards dsr
+      WHERE dsr.day = (
+        SELECT MAX(day) 
+        FROM daily_scheduled_rewards dsr2 
+        WHERE dsr2.participant_id = dsr.participant_id
+      )
```
bajtos, line 145: I don't understand why we need this more complex query. Wouldn't the following work?

```sql
      SELECT DISTINCT ON (participant_id) participant_id, scheduled_rewards
      FROM daily_scheduled_rewards
      ORDER BY participant_id, day DESC
```

bajtos (changes requested):
The high-level direction looks good to me 👍🏻 

Let's discuss the implementation details now.

bajtos:
@Hany-Almnaem I have a question about the performance.

- In #324, the current query takes 26ms to plan and 988ms to execute.
- In #324, the new query takes 2ms to plan and 580ms to execute.
- In this PR, the new query takes 44ms to plan and 1327ms to execute.

That looks like a step in the wrong direction to me.
- Is this related to https://github.com/CheckerNetwork/spark-stats/pull/340#discussion_r1990977496?
- Are we missing any indices?

Hany-Almnaem:
@bajtos You alright,
 the increased execution time is likely due to the query modifications, and the main factor could be inefficient index usage.
I'll review the indexing strategy. Let me investigate further, and I'll update the PR with improvements.

db/migrations/012.do.add-foreign-keys-and-indexes.sql
Hany-Almnaem: In daily_scheduled_rewards, the uniqueness constraint is enforced by a PRIMARY KEY (day, participant_address), not a separate index.
So, instead of just dropping an index, we need to drop the old primary key and replace it with a new one using participant_id.

For daily_reward_transfers, since the table originally used to_address instead of participant_id, we have two options:

1.  Keep participant_id for consistency across both tables.
2.  Use to_address_id to align with the table’s existing column names.

Please let me know what you think about that.

Hany-Almnaem:
Supersedes Earlier PR #324

Summary of Changes

1. Unified Table Definitions

Aligned the schema with spark-evaluate to keep everything consistent and reduce confusion.

2. Combined Migrations and Clear Documentation

Merged multiple smaller migrations into one, adding detailed comments to explain each step and keep the migration history concise.

3. Removed Old Columns

Dropped legacy columns that are no longer needed after the new schema changes.

4. Added Composite Indexes

Added  Indexes for improved query performance when filtering by day and address.

5. Updated Test Cases

Adjusted existing tests to match the new schema, ensuring they accurately reflect the latest logic and structures.

6. Optimized Query Logic in API Fetchers

Switched to using participant IDs instead of participant addresses in joins, and leveraged the new indexes for faster lookups.

With these changes, we achieve a more efficient database schema, better performance on large datasets, and clearer migration steps. 
new ANALYZE

Feel free to review and let me know if anything needs further adjustment

db/migrations/012.do.add-foreign-keys-and-indexes.sql
bajtos: > So, instead of just dropping an index, we need to drop the old primary key and replace it with a new one using participant_id.

Makes sense 👍🏻 

> For daily_reward_transfers, since the table originally used to_address instead of participant_id, we have two options:
> 
> 1. Keep participant_id for consistency across both tables.
> 2. Use to_address_id to align with the table’s existing column names.

This is a good question!

Let's use `to_adress_id` to align with the existing column names.

@juliangruber do you have any thoughts on this? IIRC, you were more involved in the discussion about the column names in the `daily_reward_transfers` table.

juliangruber: Both work for me

db/migrations/008.do.create-address-mapping.sql
Hany-Almnaem: I have added an index for this

db/migrations/008.do.create-address-mapping.sql
Hany-Almnaem: I've updated the table definition.

db/migrations/009.do.backfill-address-mapping.sql
Hany-Almnaem: I've added a second INSERT query that selects from daily_reward_transfers to ensure we capture all participant addresses.

db/migrations/012.do.add-foreign-keys-and-indexes.sql
```diff
@@ -0,0 +1,10 @@
+-- Foreign Keys
+ALTER TABLE daily_scheduled_rewards
+ADD FOREIGN KEY (participant_id) REFERENCES address_mapping(participant_id);
+
+ALTER TABLE daily_reward_transfers
+ADD FOREIGN KEY (participant_id) REFERENCES address_mapping(participant_id);
```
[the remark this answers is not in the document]
Hany-Almnaem: I've dropped the participant_address columns from both tables and added NOT NULL constraints to the participant_id columns.

db/migrations/012.do.add-foreign-keys-and-indexes.sql
Hany-Almnaem: > Let's use `to_adress_id` to align with the existing column names.

I've implemented to_address_id as suggested for better alignment with existing column names and made all the related changes in my latest commit.

observer/lib/observer.js
```diff
@@ -25,11 +25,24 @@ export const observeTransferEvents = async (pgPoolStats, ieContract, provider) =
...
+    const participantResult = await pgPoolStats.query(
+      'SELECT id FROM participants WHERE participant_address = $1',
+      [event.args.to]
+    )
+    const toAddressId = participantResult.rows[0]?.id
+
+    if (!toAddressId) {
+      console.warn('Participant not found for address:', event.args.to)
+      continue // Skip if participant is not found
+    }
```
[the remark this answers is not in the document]
bajtos: There should be a test to verify how we handle this case, please add a new one if there isn't any such test.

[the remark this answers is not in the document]
Hany-Almnaem: > Can we use the same implementation we have in spark-evaluate?

This is fixed now. I tried to match the spark-evaluate approach while keeping changes minimal to the rest of the code. I also applied the other suggestions, including (adding a test case, fixing characters, preserve the original comment ).
Let me know if any further adjustments are needed.

db/migrations/011.do.update-indexes.sql
```diff
@@ -0,0 +1,10 @@
+-- Step 1: Drop old indexes referencing participant_address
+DROP INDEX IF EXISTS daily_reward_transfers_to_address_day;
+DROP INDEX IF EXISTS idx_daily_scheduled_rewards_participant_address;
+
+-- Step 2: Create new indexes for better performance (AFTER dropping old ones)
+CREATE INDEX CONCURRENTLY idx_daily_scheduled_rewards_pid_day
```
bajtos, line 6: When I run `npm migrate` on my computer, I get the following error:

```
error: CREATE INDEX CONCURRENTLY cannot run inside a transaction block
```

I am afraid we need to split this schema migration file into three files: 1) drop old indices 2) create first index 3) create second index.

I'll apply the change myself to speed up landing of this PR.

db/migrations/008.do.create-participants-table.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE participants (
+    id SERIAL PRIMARY KEY,
+    participant_address TEXT NOT NULL UNIQUE
+);
+
+-- Add an index for fast lookups
+CREATE INDEX idx_participant_address ON participants (participant_address);
```
bajtos, line 7: This index is redundant. Because `participant_address` is `UNIQUE`, there is already an index that can be used for fast lookups.

observer/test/observer.test.js
```diff
@@ -89,7 +124,7 @@ describe('observer', () => {
       ieContractMock.queryFilter = async (eventName, fromBlock) => {
         const events = [
           { args: { to: 'address1', amount: 50 }, blockNumber: 2000 },
-          { args: { to: 'address1', amount: 50 }, blockNumber: 2000 }
+          { args: { to: 'address2', amount: 150 }, blockNumber: 2000 }
```
bajtos, line 127: We need to revert this change, as it significantly changes what is being tested in this test.

observer/test/observer.test.js
```diff
@@ -101,45 +136,55 @@ describe('observer', () => {
...
       `)
-      assert.strictEqual(rows.length, 1)
-      assert.deepStrictEqual(rows, [{
-        day: today(), to_address: 'address1', amount: '100', last_checked_block: 2000
-      }])
+      assert.strictEqual(rows.length, 2)
+      assert.deepStrictEqual(rows, [
+        { day: today(), to_address: 'address1', amount: '50', last_checked_block: 2000 },
+        { day: today(), to_address: 'address2', amount: '150', last_checked_block: 2000 }
+      ])
```
bajtos, line 148: Similarly here. I think we should simply revert this block.

observer/test/observer.test.js
```diff
@@ -101,45 +136,55 @@ describe('observer', () => {
...
 
     it('should avoid querying too old blocks', async () => {
-      providerMock.getBlockNumber = async () => 2500
+      providerMock.getBlockNumber = async () => 2000
       ieContractMock.queryFilter = async (eventName, fromBlock) => {
         const events = [
           { args: { to: 'address1', amount: 50 }, blockNumber: 400 },
           { args: { to: 'address2', amount: 150 }, blockNumber: 400 },
-          { args: { to: 'address1', amount: 250 }, blockNumber: 2000 }
+          { args: { to: 'address1', amount: 150 }, blockNumber: 2000 }
```
bajtos, line 157: Ditto, let's revert.

observer/test/observer.test.js
```diff
@@ -101,45 +136,55 @@ describe('observer', () => {
...
+        FROM daily_reward_transfers
+        LEFT JOIN participants ON daily_reward_transfers.to_address_id = participants.id
+        ORDER BY to_address_id
       `)
-      assert.strictEqual(rows.length, 1)
+      assert.strictEqual(rows.length, 2)
       assert.deepStrictEqual(rows, [
-        { day: today(), to_address: 'address1', amount: '250', last_checked_block: 2500 }
+        { day: today(), to_address: 'address1', amount: '200', last_checked_block: 2000 },
+        { day: today(), to_address: 'address2', amount: '150', last_checked_block: 2000 }
```
bajtos, line 173: Ditto, let's revert.

observer/test/platform-stats.test.js
```diff
@@ -17,10 +17,18 @@ describe('platform-stats-generator', () => {
...
+    await pgClient.query('DELETE FROM participants')
     // Run all tests inside a transaction to ensure `now()` always returns the same value
     // See https://dba.stackexchange.com/a/63549/125312
     // This avoids subtle race conditions when the tests are executed around midnight.
+
+    // Insert participants for testing (addresses mapped to IDs)
+    await pgClient.query(`
+      INSERT INTO participants (id, participant_address)
+      VALUES (1, 'address1'), (2, 'address2')
+    `)
```
bajtos, line 29: FIXME: use `mapParticipantsToIds` and save the ids generated by the database instead of using hard-coded ID values `1` and `2`.

bajtos (changes requested):
I cleaned up the first part of the patch, see the commits above. I need to take a closer look at the second part.

bajtos:
Screenshot 2025-04-11 at 7 55 09

@Hany-Almnaem please don't force-push to pull requests, it makes it more difficult to incrementally review only what's changed since the last review. We use merge commits to bring new changes from the `main` branch: `git merge main` or `[Update branch]` button in GitHub UI.

stats/test/app.test.js
```diff
@@ -371,11 +371,23 @@ describe('HTTP request handler', () => {
...
+    await pgPools.stats.query('DELETE FROM daily_scheduled_rewards')
+    await pgPools.stats.query('DELETE FROM daily_reward_transfers')
+    await pgPools.stats.query('DELETE FROM participants')
+
+    // Insert participants used by the tests below
+    await pgPools.stats.query(`
+      INSERT INTO participants (id, participant_address)
+      VALUES (1, '0x20'), (2, '0x00')
+    `)
+  })
```
bajtos, line 384: We need to use `mapParticipantsToIds` instead of hard-coded participant ids.

stats/test/platform-routes.test.js
```diff
@@ -208,6 +208,23 @@ describe('Platform Routes HTTP request handler', () => {
...
+    await pgPools.stats.query(`
+      INSERT INTO participants (id, participant_address)
+      VALUES
+        (1, 'to1'),
+        (2, 'to2'),
+        (3, 'to3'),
+        (4, 'address1'),
+        (5, 'address2'),
+        (6, 'address3')
+    `)
```
bajtos, line 225: Same here - use `mapParticipantsToIds`

bajtos: Do we actually need to prepare this mapping in advance? I would expect `givenDailyRewardTransferMetrics` to take care for that.

stats/test/app.test.js
bajtos: I recommend creating a helper that will ensure the address is mapped to an ID and then call the sql query to insert into `daily_scheduled_rewards`.

stats/test/platform-routes.test.js
```diff
@@ -274,20 +291,37 @@ describe('Platform Routes HTTP request handler', () => {
...
```
bajtos, line 308: Ditto.

I recommend creating a helper that will update both `participants` and `daily_scheduled_rewards` tables, e.g. `givenDailyScheduledRewards(day, participantAddress, scheduledRewards)`.

stats/test/platform-routes.test.js
```diff
@@ -527,14 +561,27 @@ const givenMonthlyActiveStationCount = async (pgPoolEvaluate, month, stationCoun
...
-    ON CONFLICT DO NOTHING
+  for (const transfer of transferStats) {
+    const participantResult = await pgPoolStats.query(
+      'SELECT id FROM participants WHERE participant_address = $1',
+      [transfer.toAddress]
+    )
+
+    if (participantResult.rows.length === 0) {
+      throw new Error(`Participant address ${transfer.toAddress} not found`)
+    }
```
bajtos, line 572: Let's use `mapParticipantsToIds` to map new addresses to participant ids, so that we don't have to do it manually in tests.

bajtos (changes requested):
Thank you for the updates, @Hany-Almnaem. I appreciate your perseverance!

I did a bit of cleanup in [2475264](https://github.com/CheckerNetwork/spark-stats/pull/340/commits/2475264d59822eaa239a524f7e78371b9c25b0b7) to speed things up.

I identified three more areas in tests that need improving to use `mapParticipantsToIds`, please take a look at my comments above.

Hany-Almnaem:
@bajtos  Thanks for your earlier feedback and the cleanup. Much appreciated!

I noticed the opportunity to unify the helper functions, but I decided to keep them separate for now to preserve clarity between different test contexts. Just wanted to call that out in case it comes up. happy to refactor further if needed.

bajtos (changes requested):
Can you please fix the linting errors?

Screenshot 2025-04-23 at 15 39 18

db/test-helpers.js
```diff
@@ -7,31 +7,72 @@ import { mapParticipantsToIds } from '@filecoin-station/spark-evaluate/lib/platf
...
+
+  await pgClient.query(
+    `
+    INSERT INTO daily_reward_transfers (day, to_address_id, amount, last_checked_block)
+    VALUES ($1, $2, $3, $4)
+  `,
+    [day, id, amount, lastCheckedBlock],
+  );
+};
+export { mapParticipantsToIds } from "../observer/lib/map-participants-to-ids.js";
```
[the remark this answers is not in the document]
Hany-Almnaem: Hey @bajtos, thanks for flagging that.
In the last commit, I’ve now stopped importing mapParticipantsToIds directly in tests and wrapped each use-case in its own helper:

givenDailyParticipants still uses the evaluate version under the hood

givenRewardTransfer & givenScheduledRewards wrap the stats version

Tests only pull in those descriptive helpers, and only the stats-specific mapper is re-exported from spark-stats-db/test-helpers.js. That way, there’s no longer any ambiguity about which database we’re mapping against. Let me know if you’d prefer alternate names or locations.

PR #344, opened 2025-03-13
chore: Add leaderboard preview subdomains to allowed origins

pyropy:
Adds a regular expression to the list of allowed origins to allow requests from subdomains of `leaderboard-f1i.pages.dev`.


stats/lib/app.js
```diff
@@ -31,6 +31,7 @@ export const createApp = ({
       'https://checker-draft.webflow.io',
       'https://www.checker.network',
       'https://leaderboard-f1i.pages.dev',
+      /^https:\/\/[^.]+\.leaderboard-f1i\.pages\.dev$/,
```
bajtos, line 34: Can you please add some comments to explain what these additional domains are for?

bajtos (approved):

stats/lib/app.js
pyropy: I've added a following comment: "Allow all subdomains of leaderboard-f1i.pages.dev generated by Cloudflare Pages preview deployments". Do you think it needs to be more descriptive?

